PR #33, opened 2016-10-19
Removed load.js and the ability to load plugins by name from PluginCollection

### LINKED ISSUE #32: Get rid of load.js and module utils
This has been discussed in https://github.com/ckeditor/ckeditor5/issues/349#issuecomment-254756554.

> (...) then we could totally remove the `load.js` stuff and the whole `module.js`. They only make sense in the source mode, but then you can easily use imports and you don't need loading features by strings. Side note: We have to define anyway how can you instantiate two editors with a different set of features, when using one bundle. Related tickets: ckeditor/ckeditor5-dev-bundler-rollup#1 and ckeditor/ckeditor5-core#28 and ckeditor/ckeditor5-core#27 (comment).

The point is that we implemented `load.js` to match the functionality of the previous approach, but it turns out that it can't really be used and, also, it isn't really needed.

In the source version of editor you have all the modules as separate files and you can import what you need and load things by `config.features = [ FeatureA, FeatureB ];`. In the compiled version loading by string doesn't work because all the code is bundled into one blob and the mapping between names and constructors is being lost (in the source version this mapping is represented by the file structure). 

Finally, there are evidences that we'll need to name features anyway and compile the `name<->constructor` mapping into the editor (as a config most likely).

### LINKED ISSUE #2864: Get rid of load.js and module utils
This has been discussed in https://github.com/ckeditor/ckeditor5/issues/349#issuecomment-254756554.

> (...) then we could totally remove the `load.js` stuff and the whole `module.js`. They only make sense in the source mode, but then you can easily use imports and you don't need loading features by strings. Side note: We have to define anyway how can you instantiate two editors with a different set of features, when using one bundle. Related tickets: ckeditor/ckeditor5-dev-bundler-rollup#1 and ckeditor/ckeditor5-core#28 and ckeditor/ckeditor5-core#27 (comment).

The point is that we implemented `load.js` to match the functionality of the previous approach, but it turns out that it can't really be used and, also, it isn't really needed.

In the source version of editor you have all the modules as separate files and you can import what you need and load things by `config.features = [ FeatureA, FeatureB ];`. In the compiled version loading by string doesn't work because all the code is bundled into one blob and the mapping between names and constructors is being lost (in the source version this mapping is represented by the file structure). 

Finally, there are evidences that we'll need to name features anyway and compile the `name<->constructor` mapping into the editor (as a config most likely).

Reinmar:
Fixes ckeditor/ckeditor5#2864.

The code is so much nicer now :).


Reinmar:
Please don't merge it before we fix all manual tests to use normal imports.

tests/editor/editor.js
```diff
@@ -55,124 +78,103 @@ describe( 'Editor', () => {
...
 			const editor = new Editor( {
-				features: 'A,B'
+				features: [ PluginA, PluginD ]
 			} );
 
-			expect( getPlugins( editor ) ).to.be.empty;
-
 			return editor.initPlugins().then( () => {
-				expect( getPlugins( editor ).length ).to.equal( 2 );
+				editor.plugins.get( PluginA ).init;
```
pomek: What does it do?

Reinmar: Results of my tests. I forgot to remove those lines.

Reinmar: Fixed in    7594993.

fredck:
I'm a bit confused now... how will configuration work in production now, so one can decide which plugins to load (or not load) for a specific editor instance? Can you give an example?

Reinmar:
Quoting myself from #32:

> Finally, there are evidences that we'll need to name features anyway and compile the `name<->constructor` mapping into the editor (as a config most likely).

See e.g. https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/1. The current "by name" loader (which we're removing) haven't enabled that anyway because this is a different kind of functionality – it only works in the source version. In the bundled version it must be "bundler specific" and we've already had the problem that the `load.js` functionality can't work with Webpack (although, it was a different issue – a wrong version of loader used – if CJS was picked, then it could actually work to some extent).

Anyway, the point here is that the editor infrastructure shouldn't try to implement loading code at runtime, because we'd need to accommodate all scenarios directly in the core. The code can be injected into the editor constructors using the `config.features` option and `editor.plugins.load()` method all the time, but it must happen by reference (by a plugin constructor). Let's leave the decision how this constructor is provided there to the bundler.

PS. Since there's a chance that we're going to name plugins and since we must solve https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/1, there's a chance that we'll have `config.removeFeatures` and `config.extraFeatures` just like in CKE4. And, in this case, the former will have to work by name. In other words, `config.features` and `config.extraFeatures` will provide "code" and `config.removeFeatures` will be able to disable chosen features. Or something like that, cause it's still to be defined (and I haven't yet gave it a thought).

fredck:
Just to clarify, I'm definitely talking about bundles here.

Yes, when it comes to adding new features, doing so by reference is definitely acceptable. My question was more about ways to configure bundled features so one can disable or enable them with ease.

Anyway, if I understood what you said, it sounds like it'll be covered.

Reinmar:
> My question was more about ways to configure bundled features so one can disable or enable them with ease.

So, a tl;dr is that features (as it seems now) will be named and this will happen by name through `config.removeFeatures`. Loading more features will have to happen by reference and we'll need to understand how that should happen in a specific bundler.

pomek:
@Reinmar, all features have been fixed. Now we are using everywhere class constructors instead of string names

Fixed repositories (links direct to the commit):
- [ckeditor5-basic-styles](https://github.com/ckeditor/ckeditor5-basic-styles/commit/b98de37c237bf9416236431f78bd5c8e3294a078)
- [ckeditor5-typing](https://github.com/ckeditor/ckeditor5-typing/commit/4eef3a52cfa56e4c347da858aa5e09d4c5c441cf)
- [ckeditor5-link](https://github.com/ckeditor/ckeditor5-link/commit/317c4e39e5d1826abe73fb9d62e1c6f5e56d1296)
- [ckeditor5-list](https://github.com/ckeditor/ckeditor5-list/commit/f07ec7a373014e31ef86ce8005b16a0592a1c47e)
- [ckeditor5-editor-classic](https://github.com/ckeditor/ckeditor5-editor-classic/commit/90e8167e60b7345923f3090b5837d9ea108a80b2)
- [ckeditor5-autoformat](https://github.com/ckeditor/ckeditor5-autoformat/commit/2410066f648f6d68723d2c39d11afcbfeabd5058)
- [ckeditor5-undo](https://github.com/ckeditor/ckeditor5-undo/commit/5336b78ad0666919e26f13697acb94fcfc847852)
- [ckeditor5-ui-default](https://github.com/ckeditor/ckeditor5-ui-default/commit/186168773ba199adee7c2af9465d7716b64cd0e6)

Reinmar:
LGTM. If everything passes, you can merge this PR.

pomek:
I'm waiting for Bender (checking CC).

PR #35, opened 2016-10-20
Changes param of _doExecute in toggleattributecommand to object. Adds new params.

### LINKED ISSUE #2865: Change ToggleAttributeCommand#_doExecute argument to an object
**Problem:** Cannot provide additional parameters to commands like Bold or Italic. Using object instead of just `forceValue` param will fix that.

As a part of the task, there should be added additional parameters: `batch` for grouping undo steps and `ranges` for specifying ranges to apply command to.

maxbarnas:
Fixes: ckeditor/ckeditor5#2865.


Reinmar (changes requested):
There's no test for the new option.

src/command/toggleattributecommand.js
```diff
@@ -93,16 +96,16 @@ export default class ToggleAttributeCommand extends Command {
 					selection.removeAttribute( this.attributeKey );
 				}
 			} else {
-				const ranges = getSchemaValidRanges( this.attributeKey, selection.getRanges(), document.schema );
+				const workRanges = getSchemaValidRanges( this.attributeKey, selection.getRanges(), document.schema );
```
Reinmar: Unnecessary change.

src/command/toggleattributecommand.js
```diff
@@ -93,16 +96,16 @@ export default class ToggleAttributeCommand extends Command {
...
 
 				// Keep it as one undo step.
-				const batch = document.batch();
+				const workBatch = batch || document.batch();
```
Reinmar: Let's keep the simple naming `batch`. Instead of setting the `batch` in line 85, set it once, here, but to a correct value:

``` js
const batch = options.batch || document.batch();
```

Reinmar (commented):
Minor stylistic issues.

PR #48, opened 2016-12-04
Implemented editor events and Plugin#afterInit support

### LINKED ISSUE #42: Plugins can not set data, a.k.a. editor.instanceReady
Now, the order of initialization is:

```js
editor.initPlugins()
   .then( () => editor._elementReplacer.replace( element, editor.ui.element ) )
   .then( () => editor.ui.init() )
   .then( () => editor.editing.view.attachDomRoot( editor.ui.editableElement ) )
   .then( () => editor.loadDataFromEditorElement() )
   .then( () => editor )
```

It means that plugins are `init` before the `loadDataFromEditorElement`, which calls `setData`. What if a plugin wants to `setData`?

But wait, there is more! What if a plugin wants to modify initial data or just read them? Plugins are also initialized before the `ui` is initialized. What if a plugin wants to use the UI?

So there could be an event that... the instance is ready, so plugins can execute some actions on it.

Alternatively, there could be a second method in plugins called when editor is ready. Thanks to it plugins could return a promise in this method what is not possible in events.

Another solution is to check if there is not data already loaded before calling  `loadDataFromEditorElement`. It would solve only the fist problem, but it is my favorite solution because is the least intrusive and seems to be the most correct behavior (hacks-less).

### LINKED ISSUE #2869: Plugins can not set data, a.k.a. editor.instanceReady
Now, the order of initialization is:

```js
editor.initPlugins()
   .then( () => editor._elementReplacer.replace( element, editor.ui.element ) )
   .then( () => editor.ui.init() )
   .then( () => editor.editing.view.attachDomRoot( editor.ui.editableElement ) )
   .then( () => editor.loadDataFromEditorElement() )
   .then( () => editor )
```

It means that plugins are `init` before the `loadDataFromEditorElement`, which calls `setData`. What if a plugin wants to `setData`?

But wait, there is more! What if a plugin wants to modify initial data or just read them? Plugins are also initialized before the `ui` is initialized. What if a plugin wants to use the UI?

So there could be an event that... the instance is ready, so plugins can execute some actions on it.

Alternatively, there could be a second method in plugins called when editor is ready. Thanks to it plugins could return a promise in this method what is not possible in events.

Another solution is to check if there is not data already loaded before calling  `loadDataFromEditorElement`. It would solve only the fist problem, but it is my favorite solution because is the least intrusive and seems to be the most correct behavior (hacks-less).

Reinmar:
Fixes ckeditor/ckeditor5#2869.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -17,7 +17,7 @@ import mix from '../utils/mix.js';
  */
 export default class Plugin {
 	/**
-	 * Creates a new Plugin instance. This is first step of a plugin initialization.
+	 * Creates a new Plugin instance. This is the first step of a plugin initialization.
 	 * See also {@link #init} and {@link #afterInit}.
 	 *
 	 * A plugin is always instantiated after its {@link #requires dependencies} and the
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -17,7 +17,16 @@ import mix from '../utils/mix.js';
  */
 export default class Plugin {
 	/**
-	 * Creates a new Plugin instance.
+	 * Creates a new Plugin instance. This is first step of a plugin initialization.
+	 * See also {@link #init} and {@link #afterInit}.
+	 *
+	 * A plugin is always instantiated after its {@link #requires dependencies} and the
+	 * {@link #init} and {@link #afterInit} methods are called in the same order.
+	 *
+	 * Usually, you'll want to put your plugin's initialization code in the {@link #init} method.
+	 * The constructor can be understood as "before init" and used in special cases, just like
+	 * {@link #afterInit} servers for the special "after init" scenarios (e.g. code which depends on other
+	 * plugins, but which doesn't {@link #requires explicitly require} them).
 	 *
 	 * @param {module:core/editor/editor~Editor} editor
 	 */
@@ -71,11 +80,22 @@ export default class Plugin {
 	 */
 
 	/**
-	 * Initializes the plugin.
+	 * The second stage (after plugin {@link #constructor}) of plugin initialization.
+	 * Unlike the plugin constructor this method can perform asynchronous.
+	 *
+	 * A plugin's `init()` method is called after its {@link #requires dependencies} are initialized,
+	 * so in the same order as constructors of these plugins.
 	 *
 	 * @returns {null|Promise}
 	 */
 	init() {}
+
+	/**
+	 * The third (and last) stage of plugin initialization. See also {@link #constructor} and {@link #init}.
+	 *
+	 * @returns {null|Promise}
+	 */
+	afterInit() {}
 
 	/**
 	 * Destroys the plugin.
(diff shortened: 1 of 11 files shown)


fredck:
So [my comment about this](https://github.com/ckeditor/ckeditor5-core/issues/42#issuecomment-264404108) has been ignored :(

Reinmar:
Yep, because it could not be implemented. Plugins cannot have `beforeInit`, because plugin's "existence" start with their `constructor()`. So they can't do anything before that and therefore `constuctor()` becomes the earliest point that a plugin can do something. So since we have `init()`, then `constructor()` is the before init.

fredck:
So probably I wasn't clear with my comment.

I meant to say that, considering that `beforeInit` is to be discarded, `afterInit` should be discarded as well. An event could be introduced for it (like `afterPluginsInit`), so implementors would always do things in the same way, by listening to events.

Reinmar:
OK, so I misunderstood you. I was thinking about this too, but I realised that:

* they have the constructor now anyway, so there would be events, `init()`... and the constructor executed before `init()`,
* using events is less convenient (looks worse in the code) and doesn't allow you to have asynchronous listeners,
* last but not least – I feel that doing something "after init" is more common than "before init" (I checked CKE4 and there's a difference, though small).

src/plugin.js
```diff
@@ -17,7 +17,16 @@ import mix from '../utils/mix.js';
  */
 export default class Plugin {
 	/**
-	 * Creates a new Plugin instance.
+	 * Creates a new Plugin instance. This is first step of a plugin initialization.
```
oleq: s/is first/is a first

src/plugin.js
```diff
@@ -17,7 +17,16 @@ import mix from '../utils/mix.js';
...
+	 * See also {@link #init} and {@link #afterInit}.
+	 *
+	 * A plugin is always instantiated after its {@link #requires dependencies} and the
+	 * {@link #init} and {@link #afterInit} methods are called in the same order.
+	 *
+	 * Usually, you'll want to put your plugin's initialization code in the {@link #init} method.
```
oleq, line 26: The rule of thumb is that professional documentation never speaks directly to the reader. Instead:

> Usually, plugin's initialization code lands in the {@link #init} method.

oleq: The rule of thumb is that professional documentation never speaks directly to the reader. Instead:

> Usually, plugin's initialization code lands in the {@link #init} method.

oleq (dismissed):

src/plugin.js
Reinmar: That's one of the point of views. The other is that talking directly to the reader makes it simpler and more digestible. AFAIR we've been using this form from time to time.

src/plugin.js
Reinmar: Done

src/plugin.js
oleq: It sounds unprofessional to me. But it doesn't matter at this stage, we'll have another chance to review docs in the future, that's for sure.

PR #53, opened 2016-12-19
Fix getSchemaValidRanges schema query

### LINKED ISSUE #52: getSchemaValidRanges fail after schema check algorithm was improved
Some test fail with https://github.com/ckeditor/ckeditor5-engine/pull/733.

### LINKED ISSUE #2875: getSchemaValidRanges fail after schema check algorithm was improved
Some test fail with https://github.com/ckeditor/ckeditor5-engine/pull/733.

Reinmar:
Fixes ckeditor/ckeditor5#2875. Requires https://github.com/ckeditor/ckeditor5-engine/pull/733.

I also fixed one more test which fails after https://github.com/ckeditor/ckeditor5-engine/pull/733.

### DIFF
--- a/src/command/helpers/getschemavalidranges.js
+++ b/src/command/helpers/getschemavalidranges.js
@@ -9,6 +9,7 @@
 
 import TreeWalker from '../../../engine/model/treewalker.js';
 import Range from '../../../engine/model/range.js';
+import Position from '../../../engine/model/position.js';
 
 /**
  * Walks through given array of ranges and removes parts of them that are not allowed by passed schema to have the
@@ -32,8 +33,9 @@ export default function getSchemaValidRanges( attribute, ranges, schema ) {
 
 		while ( !step.done ) {
 			const name = step.value.item.name || '$text';
+			const itemPosition = Position.createBefore( step.value.item );
 
-			if ( !schema.check( { name: name, inside: last, attributes: attribute } ) ) {
+			if ( !schema.check( { name: name, inside: itemPosition, attributes: attribute } ) ) {
 				if ( !from.isEqual( last ) ) {
 					validRanges.push( new Range( from, last ) );
 				}
--- a/tests/command/helpers/getschemavalidranges.js
+++ b/tests/command/helpers/getschemavalidranges.js
@@ -5,8 +5,9 @@
 
 import Document from 'ckeditor5/engine/model/document.js';
 import Range from 'ckeditor5/engine/model/range.js';
+import Selection from 'ckeditor5/engine/model/selection.js';
 import getSchemaValidRanges from 'ckeditor5/core/command/helpers/getschemavalidranges.js';
-import { setData } from 'ckeditor5/engine/dev-utils/model.js';
+import { setData, stringify } from 'ckeditor5/engine/dev-utils/model.js';
 
 describe( 'getSchemaValidRanges', () => {
 	const attribute = 'bold';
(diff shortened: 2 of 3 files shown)


src/command/helpers/getschemavalidranges.js
```diff
@@ -32,8 +33,9 @@ export default function getSchemaValidRanges( attribute, ranges, schema ) {
 
 		while ( !step.done ) {
 			const name = step.value.item.name || '$text';
+			const itemPosition = Position.createBefore( step.value.item );
 
-			if ( !schema.check( { name: name, inside: last, attributes: attribute } ) ) {
+			if ( !schema.check( { name: name, inside: itemPosition, attributes: attribute } ) ) {
```
scofalik, line 38: With this change maybe we can pass `ignoreElementEnd: true` to `TreeWalker` options?

tests/command/helpers/getschemavalidranges.js
```diff
@@ -21,21 +22,52 @@ describe( 'getSchemaValidRanges', () => {
...
+		schema.allow( { name: 'img', attributes: 'bold', inside: 'p' } );
+		schema.allow( { name: '$text', inside: 'img' } );
+
+		setData( document, '<p>foo<img>xxx</img>bar</p>' );
+
+		const validRanges = getSchemaValidRanges( attribute, ranges, schema );
+		const sel = new Selection();
+		sel.setRanges( validRanges );
+
+		expect( stringify( root, sel ) ).to.equal( '[<p>foo<img>]xxx[</img>bar</p>]' );
```
scofalik, line 55: I wonder what would happen if we fire `batch.setAttribute()` on a range like `[</img>bar]` or even `[</img>]`. It should not change the `img`.

On one hand, the second range seems weird, on the other hand it kinda makes perfect sense...

tests/command/helpers/getschemavalidranges.js
```diff
@@ -21,21 +22,52 @@ describe( 'getSchemaValidRanges', () => {
...
 	} );
 
+	it( 'should return unmodified ranges when attribute is allowed on each item (v2 – text is allowed in img)', () => {
+		schema.allow( { name: 'img', attributes: 'bold', inside: 'p' } );
+		schema.allow( { name: '$text', inside: 'img' } );
+
+		expect( getSchemaValidRanges( attribute, ranges, schema ) ).to.deep.equal( ranges );
+	} );
+
+	it( 'should return two ranges when attribute is not allowed in one item', () => {
```
scofalik: I think the name is wrong - it should be "when attribute is allowed on all items but not on a child in item" or smth like that.

scofalik: Okay, there is "**in** one item" (not on). Please change to inside and it will be fine.

src/command/helpers/getschemavalidranges.js
scofalik: Okay I checked it and this proposal is bad. NVM.

tests/command/helpers/getschemavalidranges.js
Reinmar: I think that nothing's gonna happen because such a range has two empty flat ranges. And the first range contains `[bar]` and `[]</img>`, so I presume it will only affect the text. Still, we'd need to check this...


EDIT: I started checking it and I realised that attr op can be applied to non-flat ranges. Later on the writer uses `range.getItems()`, though, which ignores element ends, so it seems that I was right anyway.

tests/command/helpers/getschemavalidranges.js
Reinmar: It should be "on", because it's not allowed on that "xxx" text.

Issue #67, opened 2017-02-24
Support Editor.build property


Reinmar:
As discussed in https://github.com/ckeditor/ckeditor5/issues/408.

We need a static `Editor.build` property support to be able to bundle plugins and default configuration.

The array with `this.constructor.build.plugins` need to be passed to `PluginCollection()`'s constructor as "availablePlugins". By `availablePlugins` we mean those which are available to be loaded by name.

Also the `this.constructor.build.plugins` needs to be passed to `this.plugins.load()` if `this.config.get( 'plugins' )` is not defined. Alternatively, it could be provided there by using `this.config.define( 'plugins', this.constructor.build.plugins )`. The latter seems cleaner to me.

The `this.constructor.build.config` must be passed to `this.condig.define()` too.

Of course, the entire `build` property is optional, as well as its properties.

Reinmar:
Note: There were some problems with `config.define()` and config.get()` (see https://github.com/ckeditor/ckeditor5-utils/issues/112), but AFAICS they were related to objects. Here, we're dealing with arrays (plugins and other values which we want to define in the config now are arrays), so we should be fine.

PR #70, opened 2017-02-28
T/49

### LINKED ISSUE #49: Implement options extraPlugins and removePlugins
This is a response to https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/1. The ability to adjust which plugins are loaded from a bundle is implemented in exactly the same way in CKEditor 4 and I don't see another, simpler solution.

PS. Since plugins are named now (https://github.com/ckeditor/ckeditor5-core/issues/27), this option will be easy to use.

### LINKED ISSUE #2873: Implement options extraPlugins and removePlugins
This is a response to https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/1. The ability to adjust which plugins are loaded from a bundle is implemented in exactly the same way in CKEditor 4 and I don't see another, simpler solution.

PS. Since plugins are named now (https://github.com/ckeditor/ckeditor5-core/issues/27), this option will be easy to use.

pomek:

Feature: `PluginCollection` - support for loading plugin by names. Closes ckeditor/ckeditor5#2873.

Introduced a new parameter for `PluginCollection.load()` - `removePlugins`.
The parameter is a list which defines plugins that won't be loaded.

---

### Additional information


src/editor/editor.js
```diff
@@ -45,7 +45,7 @@ export default class Editor {
 		 * @readonly
 		 * @member {module:core/plugin~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this );
+		this.plugins = new PluginCollection( this, this.config.get( 'plugins' ) );
```
Reinmar: This is not according to the spec (https://github.com/ckeditor/ckeditor5-core/issues/49#issuecomment-282299034).

available plugins !== `config.plugins`

There are no available plugins build in the editor yet because it's a separate ticket: https://github.com/ckeditor/ckeditor5-core/issues/67. So you can revert this change.

src/plugincollection.js
```diff
@@ -29,9 +30,20 @@ export default class PluginCollection {
...
+
+		/**
+		 * @protected
 		 * @member {Map} module:core/plugin~PluginCollection#_plugins
 		 */
 		this._plugins = new Map();
+
+		// Save available plugins.
+		for ( const plugin of availablePlugins ) {
+			this._availablePlugins.set( plugin.pluginName, plugin );
```
Reinmar: Plugin name is optional.

src/plugincollection.js
```diff
@@ -58,48 +70,89 @@ export default class PluginCollection {
...
+			return instantiatePlugin( PluginConstructorOrName )
 				.catch( ( err ) => {
 					/**
 					 * It was not possible to load the plugin.
 					 *
 					 * @error plugincollection-load
 					 * @param {String} plugin The name of the plugin that could not be loaded.
 					 */
-					log.error( 'plugincollection-load: It was not possible to load the plugin.', { plugin: PluginConstructor } );
+					log.error( 'plugincollection-load: It was not possible to load the plugin.', { plugin: PluginConstructorOrName } );
```
Reinmar: Do I see correctly that this error will be thrown if a plugin cannot be load because it cannot be found by its name? This should be a separate, more meaningful error.

src/plugincollection.js
```diff
@@ -58,48 +70,89 @@ export default class PluginCollection {
...
+					PluginConstructor.requires.forEach( ( RequiredPluginConstructorOrName ) => {
+						if ( removePlugins.includes( RequiredPluginConstructorOrName ) ) {
+							/**
+							 * The plugin dependency cannot be loaded because is listed in `removePlugins` options.
+							 *
+							 * @error plugincollection-instance
+							 * @param {*} plugin The dependency constructor which is meant to be loaded as a plugin.
+							 * @param {*} requiredBy The parent constructor which is meant to be loaded as a plugin.
+							 */
+							// jscs:disable maximumLineLength
```
Reinmar: Don't disable the linter. Just fix the code (split the text).

src/plugincollection.js
pomek: https://github.com/ckeditor/ckeditor5-core/blob/master/src/plugincollection.js#L89

src/plugincollection.js
```diff
@@ -29,9 +30,24 @@ export default class PluginCollection {
 
 		/**
 		 * @protected
+		 * @member {Map.<String,module:core/plugin~Plugin>} module:core/plugin~PluginCollection#_availablePlugins
```
Reinmar: The key may not be a string AFAICS below. It may be a function too.

src/plugincollection.js
```diff
@@ -110,6 +166,29 @@ export default class PluginCollection {
 			} );
 		}
 
+		function getPluginConstructor( PluginConstructorOrName ) {
+			if ( typeof PluginConstructorOrName === 'function' ) {
```
Reinmar: We use `==` after `typeof`.

src/plugincollection.js
```diff
@@ -110,6 +166,29 @@ export default class PluginCollection {
...
+				return PluginConstructorOrName;
+			}
+
+			const PluginConstructor = that._availablePlugins.get( PluginConstructorOrName );
+
+			if ( !PluginConstructor ) {
+				/**
+				 * The loaded plugin module is not available.
```
Reinmar: Update to:

```js
The plugin cannot be loaded by name.

Plugin classes need to be provided to the editor before they can be loaded by name. This is usually done by the builder.

TODO update this error with links to docs because it will be a very frequent problem.

@error plugincollection-plugin-not-found
@param {String} pluginName The name of the plugin which could not be loaded.
```

Reinmar:
OK, I checked the code more carefully and there are couple of things which should be refined here.

The implementation became too complex now, mostly because unification of what kind of plugin (string or constructor) was passed to `load()` is done too late.

1. The `plugins` array should be turned into an array of constructors ASAP. This way you'll avoid having `PluginConstructor__OrName__` all around that code.
2. There should be no error if you executed `load( [ PluginA ], [ 'NonExistentPlugin' ] )`. The error which would be thrown now makes no sense in this case. We can simply ignore names for which we can't match any plugins.
3. Don't modify the `removePlugins` variable in `load()`. Create a new one which will better tell what it is. Also, from what I can see it can be just an array of constructors. It doesn't need to contain anything more.

Besides, you had a missing code coverage (which I fixed in my last commit).

pomek:
Let's try once again :D 

image

src/plugincollection.js
```diff
@@ -86,83 +86,114 @@ export default class PluginCollection {
...
-				arr.push( getPluginConstructor( PluginConstructorOrName ) );
-			} else if ( PluginConstructorOrName.pluginName ) {
-				arr.push( PluginConstructorOrName.pluginName );
-			}
+		// In order to avoid using plugin names or constructors alternatively, we map all passed plugin as plugins constructors.
+		const pluginConstructorsToLoad = plugins
+			.map( ( item ) => getPluginConstructor( item ) )
+			.filter( ( PluginConstructor ) => !!PluginConstructor );
+
+		// Check whether a length of passed array with plugins is the same like an array with mapped plugins.
```
Reinmar: As you noticed yourself, this comment needs to be refined :D

Reinmar: Besides, I don't like splitting the filtering  which is done above from the check which is done below because it's a single piece of logic – can't find a plugin – throw.

OTOH, I can see that when using `filter()` it would be hard to reject the promise and stop the execution, so the check makes sense to be separated. So let's keep it like that, but please move it to a separate function. If you'll name this function intuitively, you won't need that comment :P.

The code in here could look like this:

```js
const missingPlugins = getMissingPlugins( plugins, pluginConstructorsToLoad );

if ( missingPlugins ) {
    return Promise.reject();
}
```

src/plugincollection.js
```diff
@@ -86,83 +86,114 @@ export default class PluginCollection {
...
+		if ( pluginConstructorsToLoad.length !== plugins.length ) {
+			// Extract names of plugins which cannot be loaded.
+			const notFoundPlugins = plugins.filter( ( item ) => !pluginConstructorsToLoad.includes( item ) );
```
Reinmar: I think that this is a wrong way of getting missing plugins. In `plugins` you may have strings and in `pluginConstructorsToLoad` you will have constructors. So checking if the former is in the latter doesn't make sense. Use `getPluginConstructor()` again.

src/plugincollection.js
```diff
@@ -99,7 +162,39 @@ export default class PluginCollection {
 				assertIsPlugin( PluginConstructor );
 
 				if ( PluginConstructor.requires ) {
-					PluginConstructor.requires.forEach( loadPlugin );
+					PluginConstructor.requires.forEach( ( RequiredPluginConstructorOrName ) => {
+						const RequiredPluginConstructor = getPluginConstructor( RequiredPluginConstructorOrName );
+
+						if ( !RequiredPluginConstructor ) {
```
Reinmar: This is unnecessary in this PR. It doesn't make sense for a to require other plugins by their names.

src/plugincollection.js
```diff
@@ -86,83 +86,114 @@ export default class PluginCollection {
...
+		const pluginConstructorsToRemove = removePlugins.map( ( item ) => getPluginConstructor( item ) )
+			.filter( ( PluginConstructor ) => !!PluginConstructor );
 
-		return Promise.all( plugins.map( loadPlugin ) )
+		return Promise.all( pluginConstructorsToLoad.map( loadPlugin ) )
 			.then( () => loaded );
 
-		function loadPlugin( PluginConstructorOrName ) {
+		function loadPlugin( PluginConstructor ) {
 			// Don't load the plugin if it cannot be loaded.
```
Reinmar: This comment is unnecessary.

src/plugincollection.js
Reinmar: PS. I meant that this may be a possible feature in the future, but it's not part of the ticket for which you're making a PR, so it shouldn't be included in it.

src/plugincollection.js
```diff
@@ -16,22 +16,43 @@ import log from '@ckeditor/ckeditor5-utils/src/log';
  */
 export default class PluginCollection {
 	/**
-	 * Creates an instance of the PluginCollection class, initializing it with a set of plugins.
+	 * Creates an instance of the PluginCollection class.
+	 * Allows loading and initializing plugins and their dependencies.
 	 *
 	 * @param {module:core/editor/editor~Editor} editor
+	 * @param {Array.<Function>} availablePlugins Plugins (constructor) which the collection will be able to use
```
Reinmar: This is optional.

src/plugincollection.js
```diff
@@ -56,23 +77,62 @@ export default class PluginCollection {
...
-	 * @param {Function[]} plugins An array of {@link module:core/plugin~Plugin plugin constructors}.
+	 * @param {Array.<Function|String>} plugins An array of {@link module:core/plugin~Plugin plugin constructors}
+	 * or {@link module:core/plugin~Plugin.pluginName plugin names}. The second option (names) work only if
+	 * `availablePlugins` were passed to the {@link #constructor}.
+	 * @param {Array.<String|Function>} removePlugins Names of plugins or plugin constructors
+	 *  which should not be loaded (despite being specified in the `plugins` array).
 	 * @returns {Promise} A promise which gets resolved once all plugins are loaded and available into the
 	 * collection.
-	 * @param {Array.<module:core/plugin~Plugin>} returns.loadedPlugins The array of loaded plugins.
+	 * @returns {Promise.<Array.<module:core/plugin~Plugin>>} returns.loadedPlugins The array of loaded plugins.
```
Reinmar: This is optional.

src/plugincollection.js
```diff
@@ -56,23 +77,62 @@ export default class PluginCollection {
...
-		return Promise.all( plugins.map( loadPlugin ) )
+		// In order to avoid using plugin names or constructors alternatively, we map all passed plugin as plugins constructors.
+		const pluginConstructorsToLoad = plugins
+			.map( ( item ) => getPluginConstructor( item ) )
+			.filter( ( PluginConstructor ) => !!PluginConstructor );
+
+		const missingPlugins = getMissingPlugins( plugins );
+
+		if ( missingPlugins ) {
+			log.error( 'plugincollection-load: It was not possible to load the plugins.', { plugins: missingPlugins } );
```
Reinmar: It should be the same error as the rejection below. A single error means one thing. Here, you tried to reuse an error defined in another place, but you changed its meaning (from one plugin to multiple plugins and from "can't be loaded due to an error" to "can't be loaded because wasn't provided").

src/plugincollection.js
```diff
@@ -56,23 +77,62 @@ export default class PluginCollection {
...
+	load( plugins, removePlugins = [] ) {
 		const that = this;
 		const editor = this._editor;
 		const loading = new Set();
 		const loaded = [];
 
-		return Promise.all( plugins.map( loadPlugin ) )
+		// In order to avoid using plugin names or constructors alternatively, we map all passed plugin as plugins constructors.
+		const pluginConstructorsToLoad = plugins
+			.map( ( item ) => getPluginConstructor( item ) )
```
Reinmar: It's not an item. It's a plugin name or constructor.

Reinmar:
Feature: Added support for loading plugins by name and the "removePlugins" option. Closes #49.

PR #71, opened 2017-03-02
Added support for `Editor.build` property

pomek:

Feature: Added support for `Editor.build` property. Closes #47.

---

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -31,6 +31,8 @@ export default class Editor {
 	 * @param {Object} config The editor config.
 	 */
 	constructor( config ) {
+		const availablePlugins = findAvailablePlugin( this.constructor.build );
+
 		/**
 		 * Holds all configurations specific to this editor instance.
 		 *
@@ -45,7 +47,7 @@ export default class Editor {
 		 * @readonly
 		 * @member {module:core/plugin~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this );
+		this.plugins = new PluginCollection( this, availablePlugins );
 
 		/**
 		 * Commands registered to the editor.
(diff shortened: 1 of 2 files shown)


src/editor/editor.js
```diff
@@ -45,7 +47,7 @@ export default class Editor {
 		 * @readonly
 		 * @member {module:core/plugin~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this );
+		this.plugins = new PluginCollection( this, availablePlugins );
```
Reinmar, line 50: This is overly complicated. All we need here is `new PluginCollection( this, this.constructor.build && this.constructor.build.plugins )`

Or, at least remove the `findAvailablePlugins` function and define `availablePlugins` as:

```js
const availablePlugins = this.constructor.build && this.constructor.build.plugins;
```

Reinmar:
This PR closes #67, not #47.

src/editor/editor.js
```diff
@@ -189,6 +193,42 @@ export default class Editor {
 
 mix( Editor, EmitterMixin );
 
+// @param {module:utils/config~Config} config
+// @param {Object} builtInConfig
+// @param {Object.<String,*>} builtInConfig.config
+// @param {Array.<module:core/plugin~Plugin>} availablePlugins
+function extendEditorConfig( config, builtInConfig, availablePlugins ) {
```
Reinmar, line 200: This looks bad. It shouldn't require so much code in this class to set the defaults. It must be `Config`'s ability to easily define the defaults.

I also need to think what should extend what here. I'll let you know once I figure this out :D.

pomek:
This PR is invalid (wrong branch name). I corrected it - https://github.com/ckeditor/ckeditor5-core/pull/72.

src/editor/editor.js
pomek: Ok. When you will be ready, put your idea to https://github.com/ckeditor/ckeditor5-core/pull/72.

PR #72, opened 2017-03-02
Feature: Added support for `Editor.build` property.

### LINKED ISSUE #67: Support Editor.build property

### LINKED ISSUE #2884: Support Editor.build property
As discussed in https://github.com/ckeditor/ckeditor5/issues/408.

We need a static `Editor.build` property support to be able to bundle plugins and default configuration.

The array with `this.constructor.build.plugins` need to be passed to `PluginCollection()`'s constructor as "availablePlugins". By `availablePlugins` we mean those which are available to be loaded by name.

Also the `this.constructor.build.plugins` needs to be passed to `this.plugins.load()` if `this.config.get( 'plugins' )` is not defined. Alternatively, it could be provided there by using `this.config.define( 'plugins', this.constructor.build.plugins )`. The latter seems cleaner to me.

The `this.constructor.build.config` must be passed to `this.condig.define()` too.

Of course, the entire `build` property is optional, as well as its properties.

pomek:

Feature: Added support for `Editor.build` property. Closes ckeditor/ckeditor5#2884.

---

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -31,13 +31,7 @@ export default class Editor {
 	 * @param {Object} config The editor config.
 	 */
 	constructor( config ) {
-		/**
-		 * An array contains plugins built-in the editor.
-		 *
-		 * @protected
-		 * @member {Array.<module:core/plugin~Plugin>}
-		 */
-		this._availablePlugins = ( this.constructor.build && this.constructor.build.plugins ) || [];
+		const availablePlugins = this.constructor.build && this.constructor.build.plugins;
 
 		/**
 		 * Holds all configurations specific to this editor instance.
@@ -47,13 +41,15 @@ export default class Editor {
 		 */
 		this.config = new Config( config, this.constructor.build && this.constructor.build.config );
 
+		this.config.define( 'plugins', availablePlugins );
+
 		/**
 		 * The plugins loaded and in use by this editor instance.
 		 *
 		 * @readonly
 		 * @member {module:core/plugin~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this, this._availablePlugins );
+		this.plugins = new PluginCollection( this, availablePlugins );
 
 		/**
 		 * Commands registered to the editor.
@@ -124,10 +120,7 @@ export default class Editor {
 			.then( () => this.fire( 'pluginsReady' ) );
 
 		function loadPlugins() {
-			const pluginsFromConfig = config.get( 'plugins' ) || [];
-			const pluginsToLoad = pluginsFromConfig.concat( that._availablePlugins );
-
-			return that.plugins.load( pluginsToLoad );
+			return that.plugins.load( config.get( 'plugins' ) || [] );
 		}
 
 		function initPlugins( loadedPlugins, method ) {
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -31,6 +31,8 @@ export default class Editor {
 	 * @param {Object} config The editor config.
 	 */
 	constructor( config ) {
+		const availablePlugins = findAvailablePlugin( this.constructor.build );
+
 		/**
 		 * Holds all configurations specific to this editor instance.
 		 *
@@ -45,7 +47,7 @@ export default class Editor {
 		 * @readonly
 		 * @member {module:core/plugin~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this );
+		this.plugins = new PluginCollection( this, availablePlugins );
 
 		/**
 		 * Commands registered to the editor.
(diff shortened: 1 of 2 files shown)


pomek:
`R-` because I did not fix:

- [x] https://github.com/ckeditor/ckeditor5-core/pull/71#discussion_r103882150
- [x] https://github.com/ckeditor/ckeditor5-core/pull/71#discussion_r103883961

Reinmar:
The code in the `Editor` class should look like this:

```
constructor( config ) {
    this.config = new Config( config, this.constructor.build && this.constructor.build.config );

    const availablePlugins = this.constructor.build && this.constructor.build.plugins;
    this.config.define( 'plugins', availablePlugins );
}
```

This is not perfect, but it's the first thing which comes to my mind and sounds okeysh.

Of course, for this to work, you need a new feature in `Config`.

pomek:
I've added prefix `[WIP]` to the title of this PR.

pomek:
@Reinmar, ready to review after merging https://github.com/ckeditor/ckeditor5-utils/pull/127.

src/editor/editor.js
```diff
@@ -116,7 +124,10 @@ export default class Editor {
 			.then( () => this.fire( 'pluginsReady' ) );
 
 		function loadPlugins() {
-			return that.plugins.load( config.get( 'plugins' ) || [] );
+			const pluginsFromConfig = config.get( 'plugins' ) || [];
+			const pluginsToLoad = pluginsFromConfig.concat( that._availablePlugins );
```
Reinmar: This doesn't follow: https://github.com/ckeditor/ckeditor5-core/issues/67 

```js
this.config.define( 'plugins', availablePlugins )
```

was supposed to be done.

pomek: Fixed in 1bf9e51.

tests/editor/editor.js
```diff
@@ -61,6 +65,30 @@ describe( 'Editor', () => {
 			expect( editor.plugins ).to.be.an.instanceof( PluginCollection );
 			expect( getPlugins( editor ) ).to.be.empty;
 		} );
+
+		it( 'should extend an editor configuration using built in config', () => {
+			Editor.build = {
+				config: {
+					'foo.a': 1,
```
Reinmar: Ekhm...

tests/editor/editor.js
```diff
@@ -244,6 +272,23 @@ describe( 'Editor', () => {
 				);
 			} );
 		} );
+
+		it( 'should load plugins built in the Editor even if the passed config is empty', () => {
```
Reinmar, line 294: Missing tests:

1. When both `build.plugins` and `config.plugins` are set.
1. That plugins defined in `build.plugins` can be loaded by names.
1. That it all works when inheriting from `Editor`.

Reinmar:
R- for missing tests. You could remove half of the changes you did in this PR and the tests we have would still pass ;).

pomek:
Ready.

tests/editor/editor.js
```diff
@@ -244,6 +290,89 @@ describe( 'Editor', () => {
...
+					expect( getPlugins( editor ).length ).to.equal( 3 );
+
+					expect( editor.plugins.get( PluginA ) ).to.be.an.instanceof( Plugin );
+					expect( editor.plugins.get( PluginB ) ).to.be.an.instanceof( Plugin );
+					expect( editor.plugins.get( PluginC ) ).to.be.an.instanceof( Plugin );
+				} );
+		} );
+
+		it( 'should load plugins inherited from the base Editor', () => {
+			Editor.build = {
```
Reinmar, line 358: It should be the opposite – the test was supposed to check whether `CustomEditor` will work if you build plugins into it. Currently, we know that it works, but I want this test to verify this for the future.

Compare: https://github.com/ckeditor/ckeditor5/issues/408#issuecomment-282282997.

Reinmar:
All fine, except one test needs to be updated. I'll handle this.

PR #77, opened 2017-04-04
T/49: Implement options extraPlugins and removePlugins

### LINKED ISSUE #49: Implement options extraPlugins and removePlugins

### LINKED ISSUE #2873: Implement options extraPlugins and removePlugins

pomek:

Fix: Ability to remove plugins built in the Editor class during plugins initialization. Closes ckeditor/ckeditor5#2873.

### DIFF
--- a/tests/editor/editor.js
+++ b/tests/editor/editor.js
@@ -437,9 +437,6 @@ describe( 'Editor', () => {
 			};
 
 			const editor = new CustomEditor( {
-				plugins: [
-					'A'
-				],
 				removePlugins: [ 'D' ]
 			} );
 
--- a/tests/editor/editor.js
+++ b/tests/editor/editor.js
@@ -399,6 +399,56 @@ describe( 'Editor', () => {
 					expect( editor.plugins.get( PluginD ) ).to.be.an.instanceof( Plugin );
 				} );
 		} );
+
+		it( 'should not load plugins specified in the config as "removePlugins"', () => {
+			const editor = new Editor( {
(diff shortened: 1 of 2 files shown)


tests/editor/editor.js
```diff
@@ -399,6 +399,56 @@ describe( 'Editor', () => {
...
+
+		it( 'should not load plugins build into Editor\'s subclass when "removePlugins" option is specified', () => {
+			class CustomEditor extends Editor {}
+
+			CustomEditor.build = {
+				plugins: [ PluginA, PluginD ]
+			};
+
+			const editor = new CustomEditor( {
+				plugins: [
```
Reinmar: These two settings mean the same so you don't know which one worked if you passed both. Since we're adding support for `removePlugins` please remove `plugins` from here.

PR #81, opened 2017-05-15
t/50: ToggleAttributeCommand should correctly determine its initial state

### LINKED ISSUE #50: ToggleAttributeCommand is not calling refreshState()
Now commands are not disabled even if schema forbids given attribute in current selection. For example when image is selected bold/italic buttons should be disabled:
output

It should probably listen to document `changesDone` event or selection `change` event and call `refreshState` then.

### LINKED ISSUE #2872: ToggleAttributeCommand is not calling refreshState()
Now commands are not disabled even if schema forbids given attribute in current selection. For example when image is selected bold/italic buttons should be disabled:
output

It should probably listen to document `changesDone` event or selection `change` event and call `refreshState` then.

oleq:

Fix: ToggleAttributeCommand should correctly determine its initial state. Closes ckeditor/ckeditor5#2872.


src/command/toggleattributecommand.js
```diff
@@ -43,12 +43,23 @@ export default class ToggleAttributeCommand extends Command {
...
+		// the correct state of the command is set initially.
+		// https://github.com/ckeditor/ckeditor5-core/issues/50
+		this.listenTo( editor.document, 'changesDone', () => {
+			this.refreshValue();
+			this.refreshState();
 		} );
 	}
 
 	/**
+	 * Updates command's {@link #value value} based on current selection.
```
Reinmar: the current

src/command/toggleattributecommand.js
```diff
@@ -43,12 +43,23 @@ export default class ToggleAttributeCommand extends Command {
 		 */
 		this.set( 'value', false );
 
-		this.listenTo( this.editor.document.selection, 'change:attribute', () => {
-			this.value = this.editor.document.selection.hasAttribute( this.attributeKey );
+		// Refresh the value and state of the command on #changesDone to make sure that
+		// the correct state of the command is set initially.
+		// https://github.com/ckeditor/ckeditor5-core/issues/50
```
Reinmar: I'd say that we don't have to link to this issue here. This should be the standard way of handling refreshing.

src/editor/editor.js
```diff
@@ -89,6 +89,14 @@ export default class Editor {
 		 */
 		this.data = new DataController( this.document );
 
+		// Refresh the state of all editor commands as soon as data is ready.
+		// https://github.com/ckeditor/ckeditor5-core/issues/50
+		this.listenTo( this, 'dataReady', () => {
```
Reinmar: Sounds like we need CommandCollection finally :)

oleq: Do you know other places where we iterate over them in a similar way?

Reinmar: We'll do so on destroy (we don't destroy commands now). So, I'd rather see these methods (`destroy()` and `refreshAll()` in the `CommandCollection`).

Reinmar:
I've just realised something – why do we need to refresh commands on `dataReady` if they are supposed to be refreshed on `changesDone`? I got confused now :D 

The only situation I can see here, where refreshing on `changesDone` would not work is if an editor creator would not load initial data (e.g. if the editor should be empty initially). But, I think this needs to be handled in the engine, so there's always initial `changesDone` fired.

Reinmar:
BTW, there's one more reason why this is a better option – notice that the solution we proposed (calling `cmd.refreshState()` on `dataReady`) does not refresh the value...

So far, not every command needs a value so `refreshValue()` was not defined as `Command`'s method. It's an optional method which child classes often define. Perhaps we could change this and have `refreshState()`, `refreshValue()` and perhaps `refresh()` which calls them both, but there's a problem that `refreshState()` is not to be overridden actually. It's the public method whilst the protected `_refreshState()` is supposed to implement all the logic. We'd need to go the same way with `refreshValue()` to be consistent... Or we may have public `refresh()` and protected `_refreshValue()` and `_refreshState()`... This may be an option.

Anyway, by making `changesDone` a reliable hook we move the duty to refresh themselves to commands. And this is better – only commands should be concerned about this.

oleq:
I removed the Editor part of the fix and created a [PR](https://github.com/ckeditor/ckeditor5-link/issues/119) in the ckeditor5-link.

Reinmar:
For the future – such issues should have some new integration tests. I think that the existing ones were a bit poor, hence, the issue.

PR #82, opened 2017-05-19
T/78: Make Plugin an interface

### LINKED ISSUE #78: Make Plugin an interface
I think we agreed that `PluginCollection` can easily become a util. For that we have https://github.com/ckeditor/ckeditor5-core/issues/57.

The second part of the puzzle is to remove the `Plugin` class and make it an interface. Interface which defines that `requires` and `pluginName` static props have special meaning and that the constructor is called with a plugin's context (which is different, depending who's the owner of the plugin collection).

This will allow us to remove `ckeditor5-core` imports from many packages and will allow defining plugins on the fly. E.g. like this:

```js
function enableMarkdownDP( editor ) {
    editor.data.processor = new MarkdownDP();
}

ClassicEditor.create( el, {
    plugins: [ ArticlePreset, enableMarkdownDP ]
} )
...
```

Yep, class constructor is just a function, so you don't even have to use ES6 classes here.

### LINKED ISSUE #2913: Make Plugin an interface
I think we agreed that `PluginCollection` can easily become a util. For that we have https://github.com/ckeditor/ckeditor5-core/issues/57.

The second part of the puzzle is to remove the `Plugin` class and make it an interface. Interface which defines that `requires` and `pluginName` static props have special meaning and that the constructor is called with a plugin's context (which is different, depending who's the owner of the plugin collection).

This will allow us to remove `ckeditor5-core` imports from many packages and will allow defining plugins on the fly. E.g. like this:

```js
function enableMarkdownDP( editor ) {
    editor.data.processor = new MarkdownDP();
}

ClassicEditor.create( el, {
    plugins: [ ArticlePreset, enableMarkdownDP ]
} )
...
```

Yep, class constructor is just a function, so you don't even have to use ES6 classes here.

pomek:

Feature: Editor's plugins do not have to extend the base Plugin class. Plugins can be simple function now. Closes ckeditor/ckeditor5#2913.

---

### Additional information

In fact, this PR does not resolve ckeditor/ckeditor5#2913 in 100% but allows converting existing plugins to simple functions.

I would mark this PR as `WIP`.

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -128,6 +128,10 @@ export default class Editor {
 
 		function initPlugins( loadedPlugins, method ) {
 			return loadedPlugins.reduce( ( promise, plugin ) => {
+				if ( !plugin[ method ] ) {
+					return promise;
+				}
+
 				return promise.then( plugin[ method ].bind( plugin ) );
 			}, Promise.resolve() );
 		}
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -9,6 +9,7 @@
 
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
+import log from '@ckeditor/ckeditor5-utils/src/log';
 
 /**
  * The base class for CKEditor plugin classes.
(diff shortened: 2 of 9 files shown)


pomek:
This PR is no longer valid (https://github.com/ckeditor/ckeditor5-core/issues/78#issuecomment-302675936).

PR #83, opened 2017-05-22
T/78: Make Plugin an interface (PluginCollection instance of check)

pomek:

Feature: `PluginCollection` handles with plugins which do not extend the base `Plugin` class. Part of ckeditor/ckeditor5#2913.


Reinmar:
This PR should be merged into #84. It doesn't make sense to review them separately.

PR #84, opened 2017-05-22
T/78: Make Plugin an interface (PluginInterface)

### LINKED ISSUE #78: Make Plugin an interface

### LINKED ISSUE #2913: Make Plugin an interface

pomek:

Feature: `PluginCollection` handles with plugins which do not extend the base `Plugin` class. Introduced `PluginInterface`. Closes ckeditor/ckeditor5#2913.

`init()`, `afterInit()` and `destroy()` methods for plugins are optional now.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -13,20 +13,14 @@ import mix from '@ckeditor/ckeditor5-utils/src/mix';
 /**
  * The base class for CKEditor plugin classes.
  *
+ * @implements module:core/plugin~PluginInterface
  * @mixes module:utils/observablemixin~ObservaleMixin
  */
 export default class Plugin {
 	/**
-	 * Creates a new Plugin instance. This is the first step of a plugin initialization.
-	 * See also {@link #init} and {@link #afterInit}.
-	 *
-	 * A plugin is always instantiated after its {@link module:core/plugin~Plugin.requires dependencies} and the
-	 * {@link #init} and {@link #afterInit} methods are called in the same order.
-	 *
-	 * Usually, you'll want to put your plugin's initialization code in the {@link #init} method.
-	 * The constructor can be understood as "before init" and used in special cases, just like
-	 * {@link #afterInit} servers for the special "after init" scenarios (e.g. code which depends on other
-	 * plugins, but which doesn't {@link module:core/plugin~Plugin.requires explicitly require} them).
+	 * Creates a new Plugin instance. This is the
+	 * {@link module:core/plugin~PluginInterface#constructor first step} of a plugin initialization. See also
+	 * {@link module:core/plugin~PluginInterface#init} and {@link module:core/plugin~PluginInterface#afterInit}.
 	 *
 	 * @param {module:core/editor/editor~Editor} editor
 	 */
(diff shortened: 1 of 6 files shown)


src/plugin.js
```diff
@@ -41,70 +35,113 @@ export default class Plugin {
...
 	 */
 	destroy() {}
 }
 
 mix( Plugin, ObservableMixin );
+
+/**
+ * The base interface for CKEditor plugins.
+ *
+ * @interface Plugin
```
ma2ciek: Shouldn't it be an `InterfacePlugin`?

src/plugin.js
```diff
@@ -41,70 +35,113 @@ export default class Plugin {
...
+ * @static
+ * @readonly
+ * @member {String|undefined} module:core/plugin~PluginInterface.pluginName
+ */
+
+/**
+ * The editor instance.
+ *
+ * @readonly
+ * @member {module:core/editor/editor~Editor} module:core/plugin~PluginInterface#editor
```
ma2ciek: You should be able to write just `@member {module:core/editor/editor~Editor} #editor`

src/plugin.js
pomek: I would call it `Plugin` and the base plugin class as `BasePlugin` (which implements the `Plugin` interface).

ma2ciek: Or just be consequent with the `Plugin` name later.

src/plugin.js
```diff
@@ -41,70 +35,113 @@ export default class Plugin {
...
+ *
+ *		export default class ImageCaption {
+ *			static get requires() {
+ *				return [ Image ];
+ *			}
+ *      }
+ *
+ * @static
+ * @readonly
+ * @member {Array.<Function>|undefined} module:core/plugin~PluginInterface.requires
```
ma2ciek, line 102: Not sure if an interface might contain static members in JSDoc

src/plugin.js
```diff
@@ -41,70 +35,113 @@ export default class Plugin {
...
+ *
+ * A plugin is always instantiated after its {@link module:core/plugin~PluginInterface.requires dependencies} and the
+ * {@link #init} and {@link #afterInit} methods are called in the same order.
+ *
+ * Usually, you'll want to put your plugin's initialization code in the {@link #init} method.
+ * The constructor can be understood as "before init" and used in special cases, just like
+ * {@link #afterInit} servers for the special "after init" scenarios (e.g. code which depends on other
+ * plugins, but which doesn't {@link module:core/plugin~PluginInterface.requires explicitly require} them).
+ *
+ * @method constructor
```
ma2ciek: This method and below methods should start with a `#`.

src/plugin.js
Reinmar: For now, JSDoc doesn't support inheriting any static members (https://github.com/jsdoc3/jsdoc/issues/1229). We have a plugin for that, fortunately.

Reinmar:
OK, so:

* #83 needs to be merged into this PR.
* `init()`, `afterInit()` and `destroy()` need to be made optional.
* we need a test that everything works if you pass a simple function as a plugin (just like in the docs).

pomek:
Why couldn't you just merge the #83 into this PR?

> we need a test that everything works if you pass a simple function as a plugin (just like in the docs).

https://github.com/ckeditor/ckeditor5-core/pull/83/files#diff-817ad8f70e1fee22c97e006f62a50aeaR201

pomek:
> init(), afterInit() and destroy() need to be made optional.

It requires changes in the Editor class. Will do it tomorrow.

pomek:
@Reinmar, I guess it's ready to review.

Reinmar:
`gulp test --files=ui,undo` throws on this branch.

pomek:
Undo works fine.

pomek:
```
gulp test --files=ui/toolbar/contextual/contextualtoolbar.js,ui/panel/balloon/balloonpanelview.js
```

also works fine :|

pomek:
And `gulp test --files=ui` also works fine.

I don't understand what happens.

Reinmar:
Some tests affect other tests. Have fun debugging ;)

pomek:
I guess the tests fail because some of the plugins call `super.destroy()` that does not exist in the base Plugin class.

Reinmar:
Makes sense. So, let's add empty `destroy()` to `Plugin`, because we plan to add it anyway soon (https://github.com/ckeditor/ckeditor5-core/issues/86).

pomek:
@Reinmar, `ui` and `undo` work fine for me. Could you confirm?

PR #89, opened 2017-06-12
Command API – take two

### LINKED ISSUE #88: Command API – take two
tl;dr: It kinda seems that commands will become plugins and that editor -> command communication will be based on events which may allow improving the general code organisation and commands' public/private code organisation.

## The setting

I needed to solve a [few remaining issues with commands](https://github.com/ckeditor/ckeditor5-core/issues/66) before starting writing any docs (commands will need to be covered even in high-level documentation). During a F2F discussion in the team, it turned out that @pjasiun has a significantly different view on the command API than what we implemented so far. Being conservative about this myself I had serious doubts because any conceptual change in this API means a lot of thinking (manual, careful refactoring). However, the idea sounds really good and we were able to imagine solutions to all problems (conceptual and technical) which we brought during the meeting.

But, let's start with some more details about what commands are.

* A command represents an action (like, "apply bold") and some state.
* Editor (even a concrete build) needs to have an open API which allows integrating (to some reasonable extent) external application with it. Right now, most of this would be done by commands – existing commands expose pretty much everything one would need to make UI for a headless editor (which boils down to creating some buttons and binding their state to our commands).
* Commands will be (together with plugins) the most utilised piece of API which contributors will use. Hence, they need to be a nice, easy to understand API.
* At the same time, commands are used as links between independent features. E.g. the autoformat plugin works by using other plugins' commands. It can do that without any specific knowledge about them. This strong dependency of commands' extensibility brings a bit of complexity to the topic (the `refreshState` event and private `_doExecute()`).

The issues that I initially noticed were:

* All commands implement `_refreshState()` method (protected, because it's only a callback to the `refreshEvent`) and public `refreshValue()` because most of the commands also have values but there's no special mechanism for this. See https://github.com/ckeditor/ckeditor5-core/issues/66.
* All commands also have `_doExecute()` and while it was meant to secure commands from being executed directly (they should be executed through `editor.execute()` which also checks the state and fires events (see https://github.com/ckeditor/ckeditor5-core/issues/22)). However, even we made the mistake to use `_doExecute()` in tests which made them slightly unrealistic. This was also a problem in CKE4 where many people skipped `editor.executeCommand()` and went straight to `editor.getCommands( 'foo' ).execute()`.
* We should introduce command collection in order to easily perform command destruction and perhaps other actions like refreshing all commands on special events.

So, from one side we want to have a nice, consistent API, with no odd separation between private and public methods because this API will be used by thousands of developers. But at the same time, we want to secure it so it's used correctly. This was the starting point to the discussion.

## Alternative proposal

@pjasiun proposed a slightly different approach towards commands which may solve the above issues and other kinds of problems which we encountered when implementing features.

**The command API should be limited to "action" + "is enabled" (whether it can be executed). Value of the command and other pieces of its state (if it has any) should be implemented separately, in forms of plugins.**

This approach has a number of interesting implications.

Command API can be simplified to the bare minimum (it will be much easier to secure a small API). So far, we wanted to enclose everything the command does inside that class. So e.g. every command constructor listens on specific editor/document events in order to refresh itself. By getting rid of the `Command` class we will need to move this code out to the plugin. We will also need to move command's value management out. This may sound like a lot of new code in already big plugins, but no one said that what currently was a command cannot now be a separate plugin.

So, we'll have plugins introducing commands. Like `UndoCommandPlugin` (to make it clear – I'm not saying that we need `CommandPlugin`). Such plugins will expose those pieces of the state which we removed together with the `Command` class. However, this will be optional – one will also be able to implement a command without defining a separate plugin which will be especially useful in closed applications.

Having plugins specialised in introducing commands means that we'll be able to have dependencies between those new command plugins. @pjasiun encountered an issue in the past that a command depended on another command but it could not enforce loading it like plugin can.

Similarly, we'll have no problems with destroying commands (detaching listeners they created) – all the listeners will be added by plugins.

We even started considering an approach in which "execute" and "refreshState" actions were also listeners. What would that cause? E.g. that unloading a plugin would remove all evidences of the command from the editor as well. This is very consistent with what I [recently proposed regarding keystrokes](https://github.com/ckeditor/ckeditor5-utils/issues/161#issuecomment-303983892). I even think that perhaps all listeners added by plugins (e.g. – conversion) should be added through the plugin so they'd be removed together with it. The downside is that one will not be able to check whether command was registered (although, one can check whether a plugin was loaded).

## Cons?

The command is now a centralised "feature" instance. The plugin was understood by me more as a glue code, although the evidence are that many plugins contain quite a lot of logic. This has two implications.

I used to think of commands as the editor's external API. If one would like to create UI for the editor, without integrating with it much, commands would give them most of the information. Now this place will be gone.

Therefore, also building UI might get more complicated. Currently, UI components are bound straight to commands (button's `isOn` is bound to command's `value` and `isEnabled` to command's `isEnabled`).

To prevent that, we should move everything we have available on commands now to the plugins which will represent them. So a plugin would have `isOn` and `value` observable properties.

## Extremum

As an extreme solution we could even consider whether commands are needed at all. If everything will be handled by plugins, then why do we need to register any "action" + "is enabled" pairs.

I feel that this may be too far, but it's worth investigating.

## What's now?

The problem we're facing now is that we begun with valid issues for which the only clean solution we figured out mean moving a lot of code around. However, I don't see how I could write documentation now without clarifying this thing because commands will be one of the first things to explain.

Therefore, I plan to work on this topic. To come up with some prototypes, check how specific commands would need to be reimplemented and see if we don't miss anything. Then, we'll need a to see how we can roll out this change without blocking all the development.

## Some prototype

This is written in 5 minutes so it's nothing definitive (especially, that I've hit a wall so it wouldn't work :D). Compare with https://github.com/ckeditor/ckeditor5-core/blob/919433775c502158ad04cc53fa1a22177397d0a6/src/command/toggleattributecommand.js. Note that pieces of this code can be replaced by a util – I wanted to show the mechanism rather than focus on code simplicity.

```js
class BoldCommand extends Plugin {
	// Doesn't require anything (e.g. the BoldEngine) so one can use it with different converters
	// without any problems.

	constructor( editor ) {
		super( editor );

		this.set( 'isEnabled', false );
		this.set( 'value', false );

		this.listenTo( editor.document, 'changesDone', () => this.refresh() );

		// Register it as a command...
		// All the code below could be wrapped by a simple util:
		// editor.commands.register( 'bold', { execute, refresh } );

		this.listenTo( editor.commands, 'execute:bold', options => this._execute( options ) );
		this.listenTo( editor.commands, 'refresh:bold', ( evt, data ) => {
			data.isEnabled = this._checkEnabled();
		} );
		// Nooooo, I can't in any reasonable way set back `data.isEnabled` to `this.isEnabled`.
		// Unless someone can find some reasonably simple idea this may prove that relying on events in
		// 100% is not a good idea.
		// Edit: Perhaps the proposed above register() could do this if it knew the "this". It would be able
		// to set the `isEnabled` property. However, is it a good idea to have it set by an external class?
	}

	refresh() {
		// I don't like this bit either.

		this.commands.refresh( 'bold' );
		this.value = this._getValue();
	}

	_execute( options ) {
		// ... See https://github.com/ckeditor/ckeditor5-core/blob/919433775c502158ad04cc53fa1a22177397d0a6/src/command/toggleattributecommand.js#L94-L121
	}

	_checkEnabled() {
		const document = this.editor.document;

		return isAttributeAllowedInSelection( this.attributeKey, document.selection, document.schema );
	}

	_getValue() {
		return this.editor.document.selection.hasAttribute( this.attributeKey );
	}
}
```

Also, the `CommandDispatcher` which will be available in `editor.commands`:

```js
class CommandDispatcher {
	constructor() {
		this.on( 'beforeExecute', ( evt, options ) => {
			this.fire( `execute:${ commandName }`, options );
		}, { priority: 'lowest' } );

		this.on( 'execute', ( evt, options ) => {
			this.fire( `afterExecute:${ commandName }`, options );
		}, { priority: 'lowest' } );
	}

	execute( commandName, option ) {
		this.fire( `beforeExecute:${ commandName }`, options );
	}

	refresh( commandName ) {
		const data = { isEnabled: true };

		this.fire( `refresh:${ commandName }`, data );
	}

	// It's missing some `isEnabled( commandName )` (the same story as the issues above).
}
```

I'll be working on this further. But I hope it's clearer now what's the proposed direction.

## Related tickets

* https://github.com/ckeditor/ckeditor5-core/issues/66 – Improve typical command.
* https://github.com/ckeditor/ckeditor5-core/issues/22 – Command: afterExecute event (and possibly beforeExecute).
* https://github.com/ckeditor/ckeditor5-core/issues/14 – Move command helpers to a better place.
* https://github.com/ckeditor/ckeditor5-autoformat/issues/7 – InlineAutoformatEngine should work with commands, not attributes.

### LINKED ISSUE #2917: Command API – take two
tl;dr: It kinda seems that commands will become plugins and that editor -> command communication will be based on events which may allow improving the general code organisation and commands' public/private code organisation.

## The setting

I needed to solve a [few remaining issues with commands](https://github.com/ckeditor/ckeditor5-core/issues/66) before starting writing any docs (commands will need to be covered even in high-level documentation). During a F2F discussion in the team, it turned out that @pjasiun has a significantly different view on the command API than what we implemented so far. Being conservative about this myself I had serious doubts because any conceptual change in this API means a lot of thinking (manual, careful refactoring). However, the idea sounds really good and we were able to imagine solutions to all problems (conceptual and technical) which we brought during the meeting.

But, let's start with some more details about what commands are.

* A command represents an action (like, "apply bold") and some state.
* Editor (even a concrete build) needs to have an open API which allows integrating (to some reasonable extent) external application with it. Right now, most of this would be done by commands – existing commands expose pretty much everything one would need to make UI for a headless editor (which boils down to creating some buttons and binding their state to our commands).
* Commands will be (together with plugins) the most utilised piece of API which contributors will use. Hence, they need to be a nice, easy to understand API.
* At the same time, commands are used as links between independent features. E.g. the autoformat plugin works by using other plugins' commands. It can do that without any specific knowledge about them. This strong dependency of commands' extensibility brings a bit of complexity to the topic (the `refreshState` event and private `_doExecute()`).

The issues that I initially noticed were:

* All commands implement `_refreshState()` method (protected, because it's only a callback to the `refreshEvent`) and public `refreshValue()` because most of the commands also have values but there's no special mechanism for this. See https://github.com/ckeditor/ckeditor5-core/issues/66.
* All commands also have `_doExecute()` and while it was meant to secure commands from being executed directly (they should be executed through `editor.execute()` which also checks the state and fires events (see https://github.com/ckeditor/ckeditor5-core/issues/22)). However, even we made the mistake to use `_doExecute()` in tests which made them slightly unrealistic. This was also a problem in CKE4 where many people skipped `editor.executeCommand()` and went straight to `editor.getCommands( 'foo' ).execute()`.
* We should introduce command collection in order to easily perform command destruction and perhaps other actions like refreshing all commands on special events.

So, from one side we want to have a nice, consistent API, with no odd separation between private and public methods because this API will be used by thousands of developers. But at the same time, we want to secure it so it's used correctly. This was the starting point to the discussion.

## Alternative proposal

@pjasiun proposed a slightly different approach towards commands which may solve the above issues and other kinds of problems which we encountered when implementing features.

**The command API should be limited to "action" + "is enabled" (whether it can be executed). Value of the command and other pieces of its state (if it has any) should be implemented separately, in forms of plugins.**

This approach has a number of interesting implications.

Command API can be simplified to the bare minimum (it will be much easier to secure a small API). So far, we wanted to enclose everything the command does inside that class. So e.g. every command constructor listens on specific editor/document events in order to refresh itself. By getting rid of the `Command` class we will need to move this code out to the plugin. We will also need to move command's value management out. This may sound like a lot of new code in already big plugins, but no one said that what currently was a command cannot now be a separate plugin.

So, we'll have plugins introducing commands. Like `UndoCommandPlugin` (to make it clear – I'm not saying that we need `CommandPlugin`). Such plugins will expose those pieces of the state which we removed together with the `Command` class. However, this will be optional – one will also be able to implement a command without defining a separate plugin which will be especially useful in closed applications.

Having plugins specialised in introducing commands means that we'll be able to have dependencies between those new command plugins. @pjasiun encountered an issue in the past that a command depended on another command but it could not enforce loading it like plugin can.

Similarly, we'll have no problems with destroying commands (detaching listeners they created) – all the listeners will be added by plugins.

We even started considering an approach in which "execute" and "refreshState" actions were also listeners. What would that cause? E.g. that unloading a plugin would remove all evidences of the command from the editor as well. This is very consistent with what I [recently proposed regarding keystrokes](https://github.com/ckeditor/ckeditor5-utils/issues/161#issuecomment-303983892). I even think that perhaps all listeners added by plugins (e.g. – conversion) should be added through the plugin so they'd be removed together with it. The downside is that one will not be able to check whether command was registered (although, one can check whether a plugin was loaded).

## Cons?

The command is now a centralised "feature" instance. The plugin was understood by me more as a glue code, although the evidence are that many plugins contain quite a lot of logic. This has two implications.

I used to think of commands as the editor's external API. If one would like to create UI for the editor, without integrating with it much, commands would give them most of the information. Now this place will be gone.

Therefore, also building UI might get more complicated. Currently, UI components are bound straight to commands (button's `isOn` is bound to command's `value` and `isEnabled` to command's `isEnabled`).

To prevent that, we should move everything we have available on commands now to the plugins which will represent them. So a plugin would have `isOn` and `value` observable properties.

## Extremum

As an extreme solution we could even consider whether commands are needed at all. If everything will be handled by plugins, then why do we need to register any "action" + "is enabled" pairs.

I feel that this may be too far, but it's worth investigating.

## What's now?

The problem we're facing now is that we begun with valid issues for which the only clean solution we figured out mean moving a lot of code around. However, I don't see how I could write documentation now without clarifying this thing because commands will be one of the first things to explain.

Therefore, I plan to work on this topic. To come up with some prototypes, check how specific commands would need to be reimplemented and see if we don't miss anything. Then, we'll need a to see how we can roll out this change without blocking all the development.

## Some prototype

This is written in 5 minutes so it's nothing definitive (especially, that I've hit a wall so it wouldn't work :D). Compare with https://github.com/ckeditor/ckeditor5-core/blob/919433775c502158ad04cc53fa1a22177397d0a6/src/command/toggleattributecommand.js. Note that pieces of this code can be replaced by a util – I wanted to show the mechanism rather than focus on code simplicity.

```js
class BoldCommand extends Plugin {
	// Doesn't require anything (e.g. the BoldEngine) so one can use it with different converters
	// without any problems.

	constructor( editor ) {
		super( editor );

		this.set( 'isEnabled', false );
		this.set( 'value', false );

		this.listenTo( editor.document, 'changesDone', () => this.refresh() );

		// Register it as a command...
		// All the code below could be wrapped by a simple util:
		// editor.commands.register( 'bold', { execute, refresh } );

		this.listenTo( editor.commands, 'execute:bold', options => this._execute( options ) );
		this.listenTo( editor.commands, 'refresh:bold', ( evt, data ) => {
			data.isEnabled = this._checkEnabled();
		} );
		// Nooooo, I can't in any reasonable way set back `data.isEnabled` to `this.isEnabled`.
		// Unless someone can find some reasonably simple idea this may prove that relying on events in
		// 100% is not a good idea.
		// Edit: Perhaps the proposed above register() could do this if it knew the "this". It would be able
		// to set the `isEnabled` property. However, is it a good idea to have it set by an external class?
	}

	refresh() {
		// I don't like this bit either.

		this.commands.refresh( 'bold' );
		this.value = this._getValue();
	}

	_execute( options ) {
		// ... See https://github.com/ckeditor/ckeditor5-core/blob/919433775c502158ad04cc53fa1a22177397d0a6/src/command/toggleattributecommand.js#L94-L121
	}

	_checkEnabled() {
		const document = this.editor.document;

		return isAttributeAllowedInSelection( this.attributeKey, document.selection, document.schema );
	}

	_getValue() {
		return this.editor.document.selection.hasAttribute( this.attributeKey );
	}
}
```

Also, the `CommandDispatcher` which will be available in `editor.commands`:

```js
class CommandDispatcher {
	constructor() {
		this.on( 'beforeExecute', ( evt, options ) => {
			this.fire( `execute:${ commandName }`, options );
		}, { priority: 'lowest' } );

		this.on( 'execute', ( evt, options ) => {
			this.fire( `afterExecute:${ commandName }`, options );
		}, { priority: 'lowest' } );
	}

	execute( commandName, option ) {
		this.fire( `beforeExecute:${ commandName }`, options );
	}

	refresh( commandName ) {
		const data = { isEnabled: true };

		this.fire( `refresh:${ commandName }`, data );
	}

	// It's missing some `isEnabled( commandName )` (the same story as the issues above).
}
```

I'll be working on this further. But I hope it's clearer now what's the proposed direction.

## Related tickets

* https://github.com/ckeditor/ckeditor5-core/issues/66 – Improve typical command.
* https://github.com/ckeditor/ckeditor5-core/issues/22 – Command: afterExecute event (and possibly beforeExecute).
* https://github.com/ckeditor/ckeditor5-core/issues/14 – Move command helpers to a better place.
* https://github.com/ckeditor/ckeditor5-autoformat/issues/7 – InlineAutoformatEngine should work with commands, not attributes.

Reinmar:

Other: The command API has been redesigned. The `Command` methods are now public and consistent. Commands can be used in a standalone mode (without the editor). The `CommandCollection` was introduced and replaced the `Map` of commands used in `editor.commands`. Closes ckeditor/ckeditor5#2917.

  Besides changes mentioned in this point and in the "Breaking changes" section, other minor changes were done:

  * `Editor#execute()` now accepts multiple command arguments.
  * `Command#value` property was standardized.

BREAKING CHANGES: The `Command`'s protected `_doExecute()` and `_checkEnabled()` methods have been replaced by public `execute()` and `refresh()` methods.

BREAKING CHANGES: The `Command`'s `refreshState` event was removed and one should use `change:isEnabled` in order to control and override its state.

BREAKING CHANGES: The `core/command/command` module has been moved to the root directory (so the `Command` class is `core/command~Command` now).

BREAKING CHANGES: The `Command#refresh()` method is now automatically called on `editor.document#changesDone`.

BREAKING CHANGES: The `editor.commands` map was replaced by a `CommandCollection` instance so `editor.commands.set()` calls need to be replaced with `editor.commands.add()`.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-utils/pull/163.

I left `ToggleAttributeCommand` in `src/command` because I'll be refactoring this bit in a separate ticket (to not make this too big).


Reinmar:
Ouch! While working on refactoring the `EnterCommand` I noticed it does this:

```js
	execute() {
		const doc = this.editor.document;
		const batch = doc.batch();

		doc.enqueueChanges( () => {
			enterBlock( this.editor.data, batch, doc.selection, doc.schema );

			this.fire( 'afterExecute', { batch } );
		} );
	}
```

I recall now why we didn't know how to implement automatic `afterExecute` firing in the past :D. But I guess I'll find some more or less straightforward solution.

Reinmar:
While refactoring all the commands I realised that:

* `isEnabled` should be set to `true` by default because there are number of commands which are always on. This can be done by implementing default `refresh()` in `Command` which will also reduce code duplication.
* `isEnabled` should be checked in `execute`'s listener automatically. Every command requires it and but it's easy to forget about this. Especially in commands which are always `true` (when you look at their code only), but which can be blocked from outside (about which you may forget).
* I was also considering whether every command should be automatically refreshed by its constructor. That might be useful... but it's not possible to do it automatically in `Command#constructor()` because at that point some properties of a child command classes will not be set yet (you can't access `this` before calling `super()`). So perhaps this would be nice but it can't be done automatically anyway.

Reinmar:
I added the mentioned changes updated the suggested commit message.

oleq:
> isEnabled should be set to true by default because there are number of commands which are always on.

Can you explain this thing a little bit? Because at first glance it feels so much like [this pitfall](http://dev.ckeditor.com/ticket/10439) in v4 and it really scares me ;-) 

I suppose you meant something like "all commands' states should be refreshed upon `editor#ready`" or something like this?

Reinmar:
> Can you explain this thing a little bit?

We're talking here about the default implementation of `refresh()`. Each command can implement its own and most of the commands do that. However, a couple of simple commands are just always enabled. And hence, to avoid unnecessary code duplication and potential issues, I implemented base `Command#refresh()` method which unconditionally sets `isEnabled` to `true`. I also changed the default `isEnabled` set by the constructor.

However, this is only the "base" code and initial state. Every command is refreshed on `changesDone` anyway, so once data is loaded, commands change their state according to their own rules or base implementation if it fits.

There are two issues, though:

* Sometimes `changesDone` may not be called (in special cases where editor starts really empty). We know that and I think it's tracked in some issue, but it's a hypothetical situation now, so let's skip it.
* But you made me realise that it'd be safer if the initial value of `isEnabled` set by the constructor was `false`. This would mean that before first `refresh()` is called you won't be able to execute any of the commands. This would make perfect sense so I'll change it.

oleq:
> And hence, to avoid unnecessary code duplication and potential issues, I implemented base Command#refresh() method which unconditionally sets isEnabled to true. I also changed the default isEnabled set by the constructor.

This is exactly what I wanted to know – that the thing is not hard coded but controlled by a generic refresh. Thanks!

src/commandcollection.js
```diff
@@ -0,0 +1,103 @@
...
+	get( commandName ) {
+		return this._commands.get( commandName );
+	}
+
+	/**
+	 * Executes a command.
+	 *
+	 * @param {String} commandName The name of the command.
+	 */
+	execute( commandName, ...args ) {
```
pjasiun, line 54: Do we get much having the shortcut:
```js
editor.commands.execute( 'bold', { forceState: true } );
```
for:
```js
editor.commands.get( 'bold' ).execute( { forceState: true } );
```

I mean, the second one is consistent with:
```js
editor.commands.get( 'bold' ).isEnabled;
editor.commands.get( 'bold' ).value;
```
If we go with the first one shouldn't we have shortcuts for the rest?

pjasiun: I mean, I like `editor.commands.get( 'bold' ).execute( { forceState: true } );` because it's clear here that `{ forceState: true }` is just a parameter of the `execute` method. Nothing more, nothing less.

pjasiun: But the question "Do we get much having the shortcut" is a real, not a rhetorical question. I also have a feeling that we may need such shortcut.

Reinmar: I implemented `CommandCollection#execute()` to keep there the error. That was meant to reduce `Editor` size. The real shortcut for the external user is `editor.execute( 'bold', { forceState: true } )`. Having `Editor#execute` and `CommandCollection#execute` was only meant to shorten the `Editor` class and make it potentially easier to reimplement from scratch.

I hope that's fine.

Reinmar:
Back on review.

src/commandcollection.js
pjasiun: Now I see, that there is `editor.execute` that make sense to keep `commands.execute` then.

Reinmar:
I also pushed all the other commands refactored in https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-core/88.

Its build fails (https://travis-ci.org/ckeditor/ckeditor5/builds/242366798?utm_source=github_status&utm_medium=notification) due to a bug in API docs validation (https://github.com/ckeditor/ckeditor5-dev/issues/229) and can be ignored for now. The tests and everything pass.

pjasiun:
I checked these changes and its R+ from me.

Reinmar:
When merging t/ckeditor5-core/88 branches to master branches of respective repositories please use this merge commit:

```
Other: Aligned the implementation to the new Command API (see https://github.com/ckeditor/ckeditor5-core/issues/88).

BREAKING CHANGES: The command API has been changed.
```

src/command.js
```diff
@@ -0,0 +1,122 @@
...
+		// By default every command is refreshed when changes are applied to the model.
+		this.listenTo( this.editor.document, 'changesDone', () => {
+			this.refresh();
+		} );
+
+		this.on( 'execute', evt => {
+			if ( !this.isEnabled ) {
+				evt.stop();
+			}
+		}, { priority: 'high' } );
```
scofalik, line 75: Why not `highest` priority?

pjasiun: I can imagine that one want to capture the `execute` event even if the command is disabled in some special cases.

src/command/toggleattributecommand.js
```diff
@@ -30,43 +31,29 @@ export default class ToggleAttributeCommand extends Command {
 		/**
 		 * Attribute that will be set by the command.
 		 *
+		 * @readonly
 		 * @member {String}
 		 */
 		this.attributeKey = attributeKey;
 
 		/**
```
scofalik, line 39: Is this description still true? It's a bit convoluted and it hides the fact that we use document selection. Maybe we should write it directly that the value is set depending on document's selection.

Reinmar: Hm... I'll be still working on the `ToggleAttributeCommand` so I'll see if I can put this differently.

src/commandcollection.js
```diff
@@ -0,0 +1,103 @@
...
+		this._commands = new Map();
+	}
+
+	/**
+	 * Registers a new command.
+	 *
+	 * @param {String} commandName The name of the command.
+	 * @param {module:core/command~Command} command
+	 */
+	add( commandName, command ) {
```
oskarwrobel, line 35: `MarkerCollection` uses `set` method in a similar case. https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/markercollection.js#L85.

Reinmar: I think it depends on a language used in a certain domain. I think that you're more likely to say "add a command to the editor" rather than "set a command". But I'd say "set a keystroke".

The are many uses of `set()` and `add()` across our code. I don't think we'll be able to justify certain picks :D

Reinmar: One more thing – in `MarkerCollection` there's `set()` because it allows to override values (it even fires `remove` and `add`). We don't expect that in `CommandCollection` for instance – it should throw (it doesn't now, but that was the assumed semantic).

PR #114, opened 2018-01-10
Cleaned up Editor classes

### LINKED ISSUE #110: Clean-up creating model, controllers, view and roots
Right now the model is created in the `Editor` class, editing view in `StandardEditor`, model root is created in a creator class (i.e. `ClassicEditor`) and view root is created in a creators UI (i.e. `ClassicEditorUI`). Mess.

First of all, in the `Editor` class model should be created with both controllers, so:

https://github.com/ckeditor/ckeditor5-core/blob/f7bc3cfe65a6a000395ca9b5242603ffd2f92e14/src/editor/standardeditor.js#L44-L45

should be moved to from the `StandardEditor` to the `Editor` class. At the same time we could move keystrokes handler to the editor class keeping the whole initialisation in the one place.

Second issue is how roots should be initialised. 

I believe that the default model root should be initialised in the `Editor` class too. More problematic is where (when) editing view root should be inited since it expect dom element to be defined. This is why I think that we could handle it a little bit different. Editing controllers could bind model roots collection (it need to become observable) and editing view roots collection, so it will be enought to create root on the model to have it in the editing view too. This way editing view will be created as soon as it should be created (see https://github.com/ckeditor/ckeditor5/issues/647).

Binding editing view root and DOM element can be done later. It means that the binding will not be needed when view root will be created (it should work fine, I hope). Editing view document have `attachDomRoot` method which can be called later.

### LINKED ISSUE #2928: Clean-up creating model, controllers, view and roots
Right now the model is created in the `Editor` class, editing view in `StandardEditor`, model root is created in a creator class (i.e. `ClassicEditor`) and view root is created in a creators UI (i.e. `ClassicEditorUI`). Mess.

First of all, in the `Editor` class model should be created with both controllers, so:

https://github.com/ckeditor/ckeditor5-core/blob/f7bc3cfe65a6a000395ca9b5242603ffd2f92e14/src/editor/standardeditor.js#L44-L45

should be moved to from the `StandardEditor` to the `Editor` class. At the same time we could move keystrokes handler to the editor class keeping the whole initialisation in the one place.

Second issue is how roots should be initialised. 

I believe that the default model root should be initialised in the `Editor` class too. More problematic is where (when) editing view root should be inited since it expect dom element to be defined. This is why I think that we could handle it a little bit different. Editing controllers could bind model roots collection (it need to become observable) and editing view roots collection, so it will be enought to create root on the model to have it in the editing view too. This way editing view will be created as soon as it should be created (see https://github.com/ckeditor/ckeditor5/issues/647).

Binding editing view root and DOM element can be done later. It means that the binding will not be needed when view root will be created (it should work fine, I hope). Editing view document have `attachDomRoot` method which can be called later.

oskarwrobel:

Fix: Moved `EditingController`, `DataController` and `EditingKeystrokeHandler` from `StandardEditor` to `Editor` class. Closes ckeditor/ckeditor5#2928.
  
---
Constelation: https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-core/110

Requires:
- https://github.com/ckeditor/ckeditor5-engine/pull/1233
- https://github.com/ckeditor/ckeditor5-editor-balloon/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-editor-classic/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-editor-inline/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-enter/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-paragraph/tree/t/ckeditor5-core/110
- https://github.com/ckeditor/ckeditor5-typing/tree/t/ckeditor5-core/110

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -8,6 +8,7 @@
  */
 
 import Config from '@ckeditor/ckeditor5-utils/src/config';
+import EditingController from '@ckeditor/ckeditor5-engine/src/controller/editingcontroller';
 import PluginCollection from '../plugincollection';
 import CommandCollection from '../commandcollection';
 import Locale from '@ckeditor/ckeditor5-utils/src/locale';
@@ -18,7 +19,13 @@ import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 
 /**
- * Class representing a basic editor. It contains a base architecture, without much additional logic.
+ * Class representing editor. It contains basic editor architecture and provides API needed by plugins
+ * like {@link module:engine/controller/editingcontroller~EditingController editing pipeline} and
+ * {@link module:engine/controller/datacontroller~DataController data pipeline}.
+ *
+ * Besides it creates main {@link module:engine/model/rootelement~RootElement} using
+ * {@link module:engine/model/document~Document#createRoot createRoot} method what automatically creates
+ * corresponding {@link module:engine/view/rooteditableelement~RootEditableElement view root} element.
  *
  * See also {@link module:core/editor/standardeditor~StandardEditor}.
  *
@@ -74,30 +81,6 @@ export default class Editor {
 		this.t = this.locale.t;
 
 		/**
-		 * The editor's model.
-		 *
-		 * The center of the editor's abstract data model.
-		 *
-		 * Besides the model, the editor usually contains two controllers –
-		 * {@link #data data controller} and {@link #editing editing controller}.
-		 * The former is used e.g. when setting or retrieving editor data and contains a useful
-		 * set of methods for operating on the content. The latter controls user input and rendering
-		 * the content for editing.
-		 *
-		 * @readonly
-		 * @member {module:engine/model/model~Model}
-		 */
-		this.model = new Model();
-
-		/**
-		 * The {@link module:engine/controller/datacontroller~DataController data controller}.
-		 *
-		 * @readonly
-		 * @member {module:engine/controller/datacontroller~DataController}
-		 */
-		this.data = new DataController( this.model );
-
-		/**
 		 * Defines whether this editor is in read-only mode.
 		 *
 		 * In read-only mode the editor {@link #commands commands} are disabled so it is not possible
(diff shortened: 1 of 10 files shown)


src/editor/editor.js
```diff
@@ -8,17 +8,26 @@
...
 /**
- * Class representing a basic editor. It contains a base architecture, without much additional logic.
+ * Class representing editor. It contains basic editor architecture and provides API needed by plugins
+ * like {@link module:engine/controller/editingcontroller~EditingController editing pipeline},
+ * {@link module:engine/controller/datacontroller~DataController data pipeline} and
+ * {module:core/editingkeystrokehandler~EditingKeystrokeHandler keystroke handler}.
+ *
+ * Besides it creates main {@link module:engine/model/rootelement~RootElement} using
+ * {@link module:engine/model/document~Document#createRoot createRoot} method what automatically creates
+ * corresponding {@link module:engine/view/rooteditableelement~RootEditableElement view root} element.
```
pjasiun: > Class representing the base of the editor. It is the API all plugins can expect to get when using `editor` property. Editors implementation (like Classic Editor or Inline Editor) should extend this class. They can add their own methods and properties.

pjasiun: I don't think we need to explain what the architecture is here (we need an article for it) and I don't think we should list properties (they will be listed anyway). Instead, I focused more what this subset of properties means.

pjasiun: You can also add "see StandardEditor".

src/editor/standardeditor.js
```diff
@@ -8,16 +8,13 @@
...
-import EditingController from '@ckeditor/ckeditor5-engine/src/controller/editingcontroller';
 import isFunction from '@ckeditor/ckeditor5-utils/src/lib/lodash/isFunction';
 
 import getDataFromElement from '@ckeditor/ckeditor5-utils/src/dom/getdatafromelement';
 import setDataInElement from '@ckeditor/ckeditor5-utils/src/dom/setdatainelement';
 
 /**
- * Class representing a typical browser-based editor. It handles a single source element and
- * uses {@link module:engine/controller/editingcontroller~EditingController}.
+ * Class provides API methods for getting and setting data.
```
pjasiun: > The standard editor class, extending base editor by adding methods needed in the typical editor implementations: editor with a single editable area created based on the DOM element. It provides base API to manage data and to integrate the editor with the form when it is initialized on the textarea element.

PR #127, opened 2018-05-10
Feature: Introduced PendingActions plugin.

### LINKED ISSUE #126: Introduce `PendingActions` plugin
We need a `PendingActions`plugin. The util plugin will aggregate information about long-lasting actions in the editor, like image upload, collaboration data synchronization or adding a comment. It needs to have 2 APIs:
 - for plugins, to add a pending action with the description, update description, 
 - for integrators, to get information if there are any pending action when I want to close the editor, get descriptions, notice when a description change.

### LINKED ISSUE #2934: Introduce `PendingActions` plugin
We need a `PendingActions`plugin. The util plugin will aggregate information about long-lasting actions in the editor, like image upload, collaboration data synchronization or adding a comment. It needs to have 2 APIs:
 - for plugins, to add a pending action with the description, update description, 
 - for integrators, to get information if there are any pending action when I want to close the editor, get descriptions, notice when a description change.

oskarwrobel:

Feature: Introduced PendingActions plugin. Closes ckeditor/ckeditor5#2934.

---

This is a very basic implementation that covers all current needs.

It is possible to add pending action with a custom message and change this message. Message is observable.

Besides plugin listens to `window#beforeunload` event and displays prompt in a case of pending action.

```javascript
const action = this.editor.plugins.get( 'PendingActions' ).add( 'Custom message' );

action.message = 'New message';

this.editor.plugins.get( 'PendingActions' ).remove( action );
```

### DIFF
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -7,8 +7,11 @@
  * @module core/pendingactions
  */
 
+/* global window */
+
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import DOMEmitterMixin from '@ckeditor/ckeditor5-utils/src/dom/emittermixin';
 import Collection from '@ckeditor/ckeditor5-utils/src/collection';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
@@ -21,6 +24,21 @@ export default class PendingActions extends Plugin {
 	 */
 	static get pluginName() {
 		return 'PendingActions';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	constructor( editor ) {
+		super( editor );
+
+		/**
+		 * DOM Emitter.
+		 *
+		 * @private
+		 * @type {module:utils/dom/emittermixin~EmitterMixin}
+		 */
+		this._domEmitter = Object.create( DOMEmitterMixin );
 	}
 
 	/**
@@ -43,6 +61,15 @@ export default class PendingActions extends Plugin {
 		 * @type {module:utils/collection~Collection}
 		 */
 		this._actions = new Collection( { idProperty: '_id' } );
+
+		// It's not possible to easy test it because karma uses `beforeunload` event
+		// to warn before full page reload and this event cannot be dispatched manually.
+		/* istanbul ignore next */
+		this._domEmitter.listenTo( window, 'beforeunload', ( evtInfo, domEvt ) => {
+			if ( this.isPending ) {
+				domEvt.returnValue = this._actions.get( 0 ).message;
+			}
+		} );
 	}
 
 	/**
@@ -81,5 +108,10 @@ export default class PendingActions extends Plugin {
 	 */
 	[ Symbol.iterator ]() {
 		return this._actions[ Symbol.iterator ]();
+	}
+
+	destroy() {
+		super.destroy();
+		this._domEmitter.stopListening();
 	}
 }
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -61,6 +61,7 @@ export default class PendingActions extends Plugin {
 		 * @type {module:utils/collection~Collection}
 		 */
 		this._actions = new Collection( { idProperty: '_id' } );
+		this._actions.delegate( 'add', 'remove' ).to( this );
 
 		// It's not possible to easy test it because karma uses `beforeunload` event
 		// to warn before full page reload and this event cannot be dispatched manually.
@@ -110,8 +111,25 @@ export default class PendingActions extends Plugin {
 		return this._actions[ Symbol.iterator ]();
 	}
 
+	/**
+	 * @inheritDoc
+	 */
 	destroy() {
 		super.destroy();
 		this._domEmitter.stopListening();
 	}
+
+	/**
+	 * Fired when an action is added to the list.
+	 *
+	 * @event add
+	 * @param {Object} action The added action.
+	 */
+
+	/**
+	 * Fired when an action is removed from the list.
+	 *
+	 * @event remove
+	 * @param {Object} action The removed action.
+	 */
 }
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -75,7 +75,7 @@ export default class PendingActions extends Plugin {
 		/* istanbul ignore next */
 		this._domEmitter.listenTo( window, 'beforeunload', ( evtInfo, domEvt ) => {
 			if ( this.isPending ) {
-				domEvt.returnValue = this._actions.get( 0 ).message;
+				domEvt.returnValue = this.first.message;
 			}
 		} );
 	}
@@ -115,6 +115,15 @@ export default class PendingActions extends Plugin {
 	remove( action ) {
 		this._actions.remove( action );
 		this.isPending = !!this._actions.length;
+	}
+
+	/**
+	 * Returns first action from the list.
+	 *
+	 * returns {Object} Pending action object.
+	 */
+	get first() {
+		return this._actions.get( 0 );
 	}
 
 	/**
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -16,6 +16,13 @@ import Collection from '@ckeditor/ckeditor5-utils/src/collection';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
 /**
+ * List of editor pending actions.
+ *
+ * Any asynchronous action that should be finished before the editor destroy (like file upload) should be registered
+ * in this plugin and removed after finish.
+ *
+ * This plugin listens to `window#beforeunload` event and displays browser prompt when a pending action is in progress.
+ *
  * @extends module:core/plugin~Plugin
  */
 export default class PendingActions extends Plugin {
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -33,7 +33,6 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  * 		const pendingActions = editor.plugins.get( 'PendingActions' );
  * 		const action = pendingActions.add( 'Unsaved changes.' );
  *
- *
  * 		pendingActions.remove( action );
  *
  * Getting pending actions:
--- a/src/pendingactions.js
+++ b/src/pendingactions.js
@@ -15,8 +15,6 @@ import DOMEmitterMixin from '@ckeditor/ckeditor5-utils/src/dom/emittermixin';
 import Collection from '@ckeditor/ckeditor5-utils/src/collection';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
-//  This plugin is not able to interrupt actions or manage that in any other way.
-
 /**
  * List of editor pending actions.
  *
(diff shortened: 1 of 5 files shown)


src/pendingactions.js
```diff
@@ -0,0 +1,145 @@
...
+	 * Adds action to the list of pending actions.
+	 *
+	 * Methods returns an object with observable message property. Message can be changed.
+	 *
+	 * @param {String} message
+	 * @returns {Object} Observable object that represents a pending action.
+	 */
+	add( message ) {
+		if ( typeof message !== 'string' ) {
+			throw new CKEditorError( 'pendingactions-add-invalid-message: Message has to be a string.' );
```
pomek: Missing docs for the `pendingactions-add-invalid-message` error.

See https://github.com/ckeditor/ckeditor5-core/blob/7e7b95007a9f92fe071d9f847536c3499d467f1e/src/plugincollection.js#L186-L197.

oskarwrobel: AFAIK there is no need to add docs to the error when an additional explanation is not necessary.

pomek: If you miss the `@error` JSDoc tag, the error will not appear on the [error codes page](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/support/error-codes.html):

image

After adding:

```diff
+/**
+ * Some docs for the error.
+ *
+ * @error pendingactions-add-invalid-message
+ */
throw new CKEditorError( 'pendingactions-add-invalid-message: Message has to be a string.' );
```

the error appears:

image

One more thing. When I built the docs, I noticed that: 

```
No errors found during the validation.
WARNING: The @type tag does not permit a description; the description will be ignored. File: pendingactions.js, line: 55
```

Should it be `@member` instead of `@type`?

oskarwrobel: Ok, we have more errors without docs and probably we should fix them too. I'll add docs to this error. And I'll change `@type` to `@member` :) Thx.

pomek: > we have more errors without docs

Would be good to verify this and report as a new ticket.

oskarwrobel: OTOH I still think that additional explanation for this error is not necessary because the error message is clear and we don't need to explain it in the docs. @pjasiun WDYT? Should we always add docs to error or only when error needs an additional explanation?

pomek: You don't need to introduce any additional description. See here – https://github.com/ckeditor/ckeditor5-engine/blob/a721f95601511a6fcbb7d43becb957af64f2d157/src/view/attributeelement.js#L109-L117

We need the proper documentation for the error in docs, so

```js
/**
 * Message has to be a string.
 *
 * @error pendingactions-add-invalid-message
 */
```

is enough.

oskarwrobel: I understand. But I'm not sure if documentation for this kind of errors is really necessary. We have more undocumented errors. Let's wait for @pjasiun or @Reinmar comment. If we decided to document all errors because we want to have them in generated docs then it's fine and I'll add this doc.

oskarwrobel: Ok, I checked and each new type of error needs docs. When the error is repeated then docs is not necessary.

src/pendingactions.js
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		const observable = Object.create( ObservableMixin );
```
pjasiun: The variable should be called `action`, instead of `observable`.

src/pendingactions.js
```diff
@@ -0,0 +1,145 @@
...
+		 */
+		this._actions = new Collection( { idProperty: '_id' } );
+		this._actions.delegate( 'add', 'remove' ).to( this );
+
+		// It's not possible to easy test it because karma uses `beforeunload` event
+		// to warn before full page reload and this event cannot be dispatched manually.
+		/* istanbul ignore next */
+		this._domEmitter.listenTo( window, 'beforeunload', ( evtInfo, domEvt ) => {
+			if ( this.isPending ) {
+				domEvt.returnValue = this._actions.get( 0 ).message;
```
pjasiun: Since getting the first action looks like a common case, maybe we should have: `this.get`, which will do `this._actions.get` or the property `first`?

pjasiun: More I think about it, more I feel that the plugin should have `Collection` API. Not everybody feels comfortable with iterators (even you used `_actions.get( 0 )`, it is more convenient). So as much as I love iterator, I think that `get` and `length` should be also provided.

tests/manual/pendingactions.js
```diff
@@ -0,0 +1,80 @@
...
+				.then( () => wait( 500 ) )
+				.then( () => ( action.message = 'Dynamic pending action 20%.' ) )
+				.then( () => wait( 500 ) )
+				.then( () => ( action.message = 'Dynamic pending action 40%.' ) )
+				.then( () => wait( 500 ) )
+				.then( () => ( action.message = 'Dynamic pending action 60%.' ) )
+				.then( () => wait( 500 ) )
+				.then( () => ( action.message = 'Dynamic pending action 80%.' ) )
+				.then( () => wait( 500 ) )
+				.then( () => ( action.message = 'Dynamic pending action 1000%.' ) )
```
pjasiun: 1000%?
That Escalated Quickly

tests/manual/pendingactions.js
```diff
@@ -0,0 +1,80 @@
...
+		window.editor = editor;
+
+		const pendingActions = editor.plugins.get( PendingActions );
+		const actionsEl = document.querySelector( '.pending-actions' );
+
+		document.querySelector( '#add-action' ).addEventListener( 'click', () => {
+			const action = pendingActions.add( 'Static pending action.' );
+
+			wait( 5000 ).then( () => pendingActions.remove( action ) );
+		} );
```
pjasiun: Do we need these "static pending actions"? Can't we have only one type of actions in this manual tests (dynamic)? More buttons in manual tests mean more things to click, to test, more code to maintain. If "dynamic" pending actions cover all cases lets make it simple.

src/pendingactions.js
```diff
@@ -0,0 +1,145 @@
...
+			}
+		} );
+	}
+
+	/**
+	 * Adds action to the list of pending actions.
+	 *
+	 * Methods returns an object with observable message property. Message can be changed.
```
pjasiun:
```
This method returns an action object with observable message property. The action object can be later used in the remove method. It also allows you to change the message.
```

pjasiun: Also, it would be useful to put an example here.

src/pendingactions.js
```diff
@@ -0,0 +1,145 @@
...
+import Plugin from './plugin';
+import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import DOMEmitterMixin from '@ckeditor/ckeditor5-utils/src/dom/emittermixin';
+import Collection from '@ckeditor/ckeditor5-utils/src/collection';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+/**
+ * List of editor pending actions.
+ *
+ * Any asynchronous action that should be finished before the editor destroy (like file upload) should be registered
```
pjasiun: > should be finished before the editor destroy

This plugin has nothing to do with editing destroy. It is to handle long-lasting actions. To synchronise plugins which execute such action (i.e. file upload) and the editor integration. It gives a developer, who integrates the editor, an easy way to check if there are any pending action whenever such information is needed. All plugins, which register pending action provides also a message what action is ongoing which can be displayed to a user and let him decide if he wants to interrupt the action or wait. This plugin is not able to interrupt actions or manage that in any other way.

src/pendingactions.js
oskarwrobel: Done.

src/pendingactions.js
oskarwrobel: Done.

src/pendingactions.js
oskarwrobel: In f2f talk we decided to add getter `first` for getting first pending action from the list.

tests/manual/pendingactions.js
oskarwrobel: Done.

tests/manual/pendingactions.js
oskarwrobel: Done.

src/pendingactions.js
```diff
@@ -0,0 +1,178 @@
...
+		this._actions.delegate( 'add', 'remove' ).to( this );
+
+		// It's not possible to easy test it because karma uses `beforeunload` event
+		// to warn before full page reload and this event cannot be dispatched manually.
+		/* istanbul ignore next */
+		this._domEmitter.listenTo( window, 'beforeunload', ( evtInfo, domEvt ) => {
+			if ( this.isPending ) {
+				domEvt.returnValue = this.first.message;
+			}
+		} );
```
pjasiun: We talked with @Reinmar and agreed that this logic is not needed with no autosave plugin so we should move it there. If one does not use autosave plugins, he also does not care about closing browser with pending action. Content is not saved anyway.

oskarwrobel: Done.

src/pendingactions.js
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * List of editor pending actions.
+ *
+ * This plugin should be used to synchronise plugins that execute long-lasting actions
+ * (i.e. file upload) with the editor integration. It gives a developer, who integrates the editor,
+ * an easy way to check if there are any pending action whenever such information is needed.
+ * All plugins, which register pending action provides also a message what action is ongoing
+ * which can be displayed to a user and let him decide if he wants to interrupt the action or wait.
+ *
+ * This plugin listens to `window#beforeunload` event and displays browser prompt when a pending action is in progress.
```
pjasiun: Not anymore.

src/pendingactions.js
```diff
@@ -0,0 +1,146 @@
...
+	/**
+	 * Adds action to the list of pending actions.
+	 *
+	 * This method returns an action object with observable message property.
+	 * The action object can be later used in the remove method. It also allows you to change the message.
+	 *
+	 *		const pendingActions = editor.plugins.get( 'PendingActions' );
+	 * 		const action = pendingActions.add( 'Upload in progress 0%' );
+	 *
+	 * 		action.message = 'Upload in progress 10%';
```
pjasiun: There is no example how to remove action.

src/pendingactions.js
```diff
@@ -0,0 +1,146 @@
...
+	 *
+	 * @param {String} message
+	 * @returns {Object} Observable object that represents a pending action.
```
pjasiun: It would be great to describe properties.

src/pendingactions.js
```diff
@@ -0,0 +1,146 @@
...
+	/**
+	 * Returns first action from the list.
+	 *
+	 * 		const pendingActions = editor.plugins.get( 'PendingActions' );
+	 *
+	 * 		pendingActions.add( 'Action 1' );
+	 * 		pendingActions.add( 'Action 2' );
+	 *
+	 *		pendingActions.first // Returns 'Action 1'
+	 *		Array.from( pendingActions ) // Returns [ 'Action 1', 'Action 2' ]
```
pjasiun: Since this sample shows usage of not only this method by also iterator interface I would move it to the class description.

src/pendingactions.js
pjasiun: Also, I would move the whole sample to the plugin description.

tests/manual/pendingactions.md
```diff
@@ -0,0 +1,6 @@
+## Pending actions
+
+1. Click `Add pending actions`
+2. Check if action message is changing (action should disappear when progress reaches 100%)
+3. Try to reload page or close the browser tab when pending action is displayed, you should see a prompt
```
pjasiun, line 5: It does not make sense now. We could have some "finish" button to check if it works.

pjasiun: Or you can add `onbeforeunload ` integration to the sample.

PR #128, opened 2018-05-28
Internal: Used `beforeChange` instead of `change` to force command read-only mode.

oskarwrobel:

Internal: Used `beforeChange` instead of `change` to force `Command` read-only mode. Closes https://github.com/ckeditor/ckeditor5-utils/issues/171.

---

### Requires
- https://github.com/ckeditor/ckeditor5-utils/pull/241

### DIFF
--- a/src/command.js
+++ b/src/command.js
@@ -79,19 +79,13 @@ export default class Command {
 		// By default commands are disabled when the editor is in read-only mode.
 		this.listenTo( editor, 'change:isReadOnly', ( evt, name, value ) => {
 			if ( value ) {
-				// See a ticket about overriding observable properties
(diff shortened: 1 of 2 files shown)


oleq (changes requested):
The change requires tests ;-)

oskarwrobel:
This functionality is tested https://github.com/ckeditor/ckeditor5-core/blob/6f55c206ffa51af11f14dc1eee00b284d81e1c87/tests/command.js#L69-L100.

Reinmar:
So why is this change needed? If everything is fine here, then there's no need for a PR. But we know that something is wrong. So let's test it.

oskarwrobel:
Ok, test is added.

PR #129, opened 2018-06-13
Changes in EditorWithUI and ElementApi interfaces

### LINKED ISSUE #64: How to create editor without replacing an existing element
@oskarwrobel asked me today how to create the classic editor with data which he stores in JS. Well, turns out that you need to create a fork of the `ClassicEditor`.

We've never thought about this enough I think. The assumption that the editor is replacing an existing element can't be generalised. It may be true in sth like 75%, making the editor hard to use in 25% cases.

We need to design the creators in a way that you'll be able to either replace an existing element or create the editor with your own data.

First of all, this means that the concept of "editor element" needs to be revisited. An editor may not have an element if it didn't replace any. So the `StandardEditor#element` property and `updateEditorElement/loadDataFromEditorElement()` methods will be optional and perhaps should be removed totally. 

Second, we need to define a more general `create()` method. The first idea which came to my mind was that create() could have two modes – if called with an element it would replace it, if called with a string, it would create an editor with that data.

HOWEVER, I'm a bit worried that we cannot allow creating an editor detached from the DOM. I'm not entirely sure that it's going to work correctly. From my past experience, it may through weird errors.

### LINKED ISSUE #65: Lack of initial paragraph when create editor with no content
feb-13-2017 12-57-22

It breaks the editor. When selection is in the first line pressing <kbd>Enter</kbd> do nothing but pressing <kbd>Backspace</kbd> creates new line.

I don't know in which repo this ticket should be created? `ckeditor5-engine`, `ckeditor5-paragraph`, `ckeditor5-typing` or other?

### LINKED ISSUE #2882: How to create editor without replacing an existing element
@oskarwrobel asked me today how to create the classic editor with data which he stores in JS. Well, turns out that you need to create a fork of the `ClassicEditor`.

We've never thought about this enough I think. The assumption that the editor is replacing an existing element can't be generalised. It may be true in sth like 75%, making the editor hard to use in 25% cases.

We need to design the creators in a way that you'll be able to either replace an existing element or create the editor with your own data.

First of all, this means that the concept of "editor element" needs to be revisited. An editor may not have an element if it didn't replace any. So the `StandardEditor#element` property and `updateEditorElement/loadDataFromEditorElement()` methods will be optional and perhaps should be removed totally. 

Second, we need to define a more general `create()` method. The first idea which came to my mind was that create() could have two modes – if called with an element it would replace it, if called with a string, it would create an editor with that data.

HOWEVER, I'm a bit worried that we cannot allow creating an editor detached from the DOM. I'm not entirely sure that it's going to work correctly. From my past experience, it may through weird errors.

pomek:
Renamed editor.element to editor.sourceElement. Introduced an "element" property to "EditorWithUI" interface.

Feature: Introduced the `#element` property to the `EditorWithUI` interface. The `#element` property from`ElementApi` interface has been renamed to `#sourceElement`. Closes ckeditor/ckeditor5#2882.

BREAKING CHANGE: Renamed attribute in `ElementApi` interface – from `#element` to `#sourceElement`.

BREAKING CHANGE: Renamed method name in `ElementApi` interface – from `updateElement()` to `updateSourceElement()`.

---

Required for:
- https://github.com/ckeditor/ckeditor5-editor-classic/pull/73
- https://github.com/ckeditor/ckeditor5-editor-inline/pull/38
- https://github.com/ckeditor/ckeditor5-editor-balloon/pull/19
- https://github.com/ckeditor/ckeditor5-build-classic/pull/47
- https://github.com/ckeditor/ckeditor5-ui/pull/409
- https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/18

~CI: https://travis-ci.org/ckeditor/ckeditor5/builds/391732850 (https://github.com/ckeditor/ckeditor5/commit/960395ee595c462518ea3ad50be67be6e3b0a2b8)~
~CI: https://travis-ci.org/ckeditor/ckeditor5/builds/398155989 (https://github.com/ckeditor/ckeditor5/commit/220950d747cfdd8e35293e63fab3964c6de14b87)~
CI: https://travis-ci.org/ckeditor/ckeditor5/builds/399007024 (https://github.com/ckeditor/ckeditor5/commit/ed334a64213a1db2ddcc8e0c60a321a2981fe2cf)


src/editor/utils/elementapimixin.js
```diff
@@ -19,8 +19,8 @@ const ElementApiMixin = {
 	/**
 	 * @inheritDoc
 	 */
-	updateElement() {
```
Reinmar, line 22: Let's add a check here that `sourceElement` exists and throw otherwise. We expect that it may not be set.

pomek: Done in e8cc7ed.

src/editor/utils/elementapimixin.js
```diff
@@ -19,8 +20,17 @@ const ElementApiMixin = {
...
-		setDataInElement( this.element, this.data.get() );
+	updateSourceElement() {
+		if ( !this.sourceElement ) {
+			/**
+			 * The `sourceElement` is required by the `ElementApi` interface.
+			 *
+			 * @error elementapimixin-missing-sourceelement
+			 */
+			throw new CKEditorError( 'elementapi-missing-sourceelement: The "sourceElement" is required by the "ElementApi" interface.' );
```
Reinmar: How's an external developer supposed to understand this? He/she doesn't know our interfaces. They know what method they used. They should learn that it doesn't work if the editor wasn't created by passing the source element.

PS. Keep the error msg short. Explain it more thoroughly in the comment above.

pomek: I improved a little the error description but I'm not able to check it out so I'm not sure whether it works... Could you help me? :D

pomek:
One more PR that requires these changes: https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/18

I missed it. The description of this PR has been updated.

PR #132, opened 2018-06-27
Added #state observable to the Editor class.

### LINKED ISSUE #124: Introduce editor.status (or state)
We should introduce `editor.status` observable property, similar to the CKEditor 4: https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#property-status

I propose to have the following possible status values:
 - initializing - during initialization chain.
 - ready - after initializing chain, the editor is ready to work.
 - destroyed - after destroy, editor and all plugins are destroyed.

This is important because there is no way to stop promises when the editor is destroyed and some plugins should check the editor status in `then`.

It can be alternatively called `state`, both `state` and `status` are fine for me. cc @Reinmar

### LINKED ISSUE #2935: Introduce editor.status (or state)
We should introduce `editor.status` observable property, similar to the CKEditor 4: https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#property-status

I propose to have the following possible status values:
 - initializing - during initialization chain.
 - ready - after initializing chain, the editor is ready to work.
 - destroyed - after destroy, editor and all plugins are destroyed.

This is important because there is no way to stop promises when the editor is destroyed and some plugins should check the editor status in `then`.

It can be alternatively called `state`, both `state` and `status` are fine for me. cc @Reinmar

oskarwrobel:

Feature: Added `#state` observable to the `Editor` class. Closes ckeditor/ckeditor5#2935.

---

~### Requires~
~- https://github.com/ckeditor/ckeditor5-editor-classic/tree/t/ckeditor5-core/124~
~- https://github.com/ckeditor/ckeditor5-editor-balloon/tree/t/ckeditor5-core/124~
~- https://github.com/ckeditor/ckeditor5-editor-inline/tree/t/ckeditor5-core/124~
~- https://github.com/ckeditor/ckeditor5-editor-decoupled/tree/t/ckeditor5-core/124~

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -119,6 +119,8 @@ export default class Editor {
 		 * @member {'initializing'|'ready'|'destroyed'} #state
 		 */
 		this.set( 'state', 'initializing' );
+		this.once( 'ready', () => ( this.state = 'ready' ), { priority: 'high' } );
+		this.once( 'destroy', () => ( this.state = 'destroyed' ), { priority: 'high' } );
 
 		/**
 		 * Defines whether this editor is in read-only mode.
@@ -241,7 +243,6 @@ export default class Editor {
 	 * @returns {Promise} A promise that resolves once the editor instance is fully destroyed.
 	 */
 	destroy() {
-		this.state = 'destroyed';
 		this.fire( 'destroy' );
 
 		this.stopListening();
@@ -287,7 +288,6 @@ export default class Editor {
 				editor.initPlugins()
 					.then( () => {
 						editor.fire( 'dataReady' );
-						editor.state = 'ready';
 						editor.fire( 'ready' );
 					} )
 					.then( () => editor )
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -106,11 +106,10 @@ export default class Editor {
 
 		/**
 		 * Indicates editor initialization status.
-		 *
 		 * The following statuses are available:
-		 * * initializing -  during initialization chain
-		 * * ready - after initializing chain, the editor is ready to work
-		 * * destroyed - after destroy, editor and all plugins are destroyed
+		 * * initializing -  during the editor initialization
+		 * * ready - after the editor initialization (editor is ready to work)
+		 * * destroyed - after the {@link #destroy editor destroy} method was called
 		 *
 		 * @observable
 		 * @member {'initializing'|'ready'|'destroyed'} #state
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -105,6 +105,19 @@ export default class Editor {
 		this.t = this.locale.t;
 
 		/**
+		 * Indicates editor initialization status.
+		 *
+		 * The following statuses are available:
+		 * * initializing -  during initialization chain
+		 * * ready - after initializing chain, the editor is ready to work
+		 * * destroyed - after destroy, editor and all plugins are destroyed
+		 *
+		 * @observable
+		 * @member {'initializing'|'ready'|'destroyed'} #state
+		 */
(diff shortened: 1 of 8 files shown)


src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
 		 */
 		this.t = this.locale.t;
 
+		/**
+		 * Indicates editor initialization status.
+		 *
+		 * The following statuses are available:
```
Reinmar: Blank line after a paragraph

src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
...
+		 * * initializing -  during initialization chain
```
Reinmar: A person reading this documentation will not know that `create()` has a chain of promises inside it. Use the terms which the audience know.

src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
...
+		 * * ready - after initializing chain, the editor is ready to work
```
Reinmar: Relate to things which the user knows. When is the editor ready? When the promise returned by `create()` is resolved.

src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
...
+		 * * destroyed - after destroy, editor and all plugins are destroyed
```
Reinmar: after the `editor.destroy()` method was called

pjasiun:
I think that specific editors should not need to set state. Base editor class should listen on `dataReady` and update it automatically. It could also change in the `destroy` event listener to make it consistent.

oskarwrobel:
> I think that specific editors should not need to set state. Base editor class should listen on dataReady and update it automatically. It could also change in the destroy event listener to make it consistent.

I was thinking about it but I assumed that you are not gonna like it :D

pjasiun:
I was also not sure if it should be `state` or `status` ([like in CKEditor 4](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#property-status)), but with @Reinmar we agreed that `state` is better.

src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
 		 */
 		this.t = this.locale.t;
 
+		/**
+		 * Indicates editor initialization status.
+		 * The following statuses are available:
```
Reinmar: Still bad. MIssing blank lines before and after this paragraph.

src/editor/editor.js
```diff
@@ -104,6 +104,19 @@ export default class Editor {
...
+		 * * initializing - during the editor initialization
+		 * * ready - after the promise returned by the {@link module:core/editor/editor~Editor.create editor create()}
+		 * method is resolved
+		 * * destroyed - after the {@link #destroy editor destroy()} method was called
```
Reinmar:
```
destroyed - after the {@link #destroy `editor.destroy()`} method was called.
```

!!!

Reinmar:
I'm going to fix the docs ;>

src/editor/editor.js
Reinmar: BTW, states aren't available. You don't choose one. The editor may be in one of the following states.

Reinmar: Actually – "may be" is not completely correct to :D "The editor is..."

oskarwrobel: That part comes from CKE4 docs :D

Reinmar:
Hehe :) I wanted to comment that this doesn't seem to be a good change if we need to make it in all repos. I can see you limited it to the `Editor` class, which is great.

PR #157, opened 2019-01-24
Rename `editor#pluginsReady` event as `editor.plugins#ready`

f1ames:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Other: Introduced `editor.plugins#ready` event. Closes ckeditor/ckeditor5#1477.

BREAKING CHANGE: The `editor#pluginsReady` event was removed. The `editor.plugins#ready` event has been introduced and should be used instead.

---

### Additional information

See https://github.com/ckeditor/ckeditor5/issues/1477 for more details. This PR mostly covers renaming `pluginsReady` event (but also necessary adjustments to new `data#ready` event). The second PR (renaming `dataReady` event) is https://github.com/ckeditor/ckeditor5-engine/pull/1646.

Created `ckeditor5` branch with all the changes - https://github.com/ckeditor/ckeditor5/tree/t/1477.

The list of related PRs in other repos which needed adjustments:
* `ckeditor5-autosave` - https://github.com/ckeditor/ckeditor5-autosave/pull/22
* `ckeditor5-editor-balloon` - https://github.com/ckeditor/ckeditor5-editor-balloon/pull/27
* `ckeditor5-editor-classic` - https://github.com/ckeditor/ckeditor5-editor-classic/pull/80
* `ckeditor5-editor-decoupled` - https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/26
* `ckeditor5-editor-inline` - https://github.com/ckeditor/ckeditor5-editor-inline/pull/46
* `ckeditor5-engine` - https://github.com/ckeditor/ckeditor5-engine/pull/1646
* `ckeditor5-ui` - https://github.com/ckeditor/ckeditor5-ui/pull/464
* `ckeditor5-paragraph` - https://github.com/ckeditor/ckeditor5-paragraph/pull/40

PR #159, opened 2019-01-29
Docs: Fixed invalid links

f1ames:

Docs: Fixed invalid links. Closes ckeditor/ckeditor5#1487.

---

### Additional information

Some leftovers after https://github.com/ckeditor/ckeditor5-core/pull/157.

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -283,10 +283,10 @@ export default class Editor {
 	 *
 	 * See the `create()` methods of the existing editor types to learn how to use them:
 	 *
-	 * * {@link module:editor-classic/classiceditor~ClassicEditor.create `ClassicEditor.create`}
-	 * * {@link module:editor-balloon/ballooneditor~BalloonEditor.create `BalloonEditor.create`}
-	 * * {@link module:editor-decoupled/decouplededitor~DecoupledEditor.create `DecoupledEditor.create`}
-	 * * {@link module:editor-inline/inlineeditor~InlineEditor.create `InlineEditor.create`}
+	 * * {@link module:editor-classic/classiceditor~ClassicEditor.create `ClassicEditor.create()`}
+	 * * {@link module:editor-balloon/ballooneditor~BalloonEditor.create `BalloonEditor.create()`}
+	 * * {@link module:editor-decoupled/decouplededitor~DecoupledEditor.create `DecoupledEditor.create()`}
+	 * * {@link module:editor-inline/inlineeditor~InlineEditor.create `InlineEditor.create()`}
 	 *
 	 * @abstract
 	 * @method module:core/editor/editor~Editor.create
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -279,17 +279,6 @@ export default class Editor {
 	/**
 	 * Creates and initializes a new editor instance.
 	 *
-	 * @static
-	 * @method module:core/editor/editor~Editor.create
-	 * @param {Object} config The editor config. You can find the list of config options in
-	 * {@link module:core/editor/editorconfig~EditorConfig}.
-	 * @returns {Promise} Promise resolved once editor is ready.
-	 * @returns {module:core/editor/editor~Editor} return.editor The editor instance.
-	 */
-
-	/**
-	 * Creates and initializes a new editor instance.
-	 *
 	 * This is an abstract method. Every editor type needs to implement its own initialization logic.
 	 *
 	 * See the `create()` methods of the existing editor types to learn how to use them:
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -49,6 +49,8 @@ export default class Editor {
 	/**
 	 * Creates a new instance of the Editor class.
 	 *
+	 * Usually, not to be used directly. See the static {@link module:core/editor/editor~Editor.create `create()`} method.
+	 *
 	 * @param {Object} [config] The editor config.
 	 */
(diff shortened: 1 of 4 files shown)


src/editor/editor.js
```diff
@@ -275,13 +275,24 @@ export default class Editor {
...
+	/**
+	 * Creates and initializes a new editor instance.
+	 *
+	 * @static
+	 * @method module:core/editor/editor~Editor.create
+	 * @param {Object} config The editor config. You can find the list of config options in
+	 * {@link module:core/editor/editorconfig~EditorConfig}.
+	 * @returns {Promise} Promise resolved once editor is ready.
+	 * @returns {module:core/editor/editor~Editor} return.editor The editor instance.
+	 */
```
pjasiun, line 293: Do we need this doc if the method does not exist anymore?

f1ames: It exists in subclasses (but has own docs there) but also `editor#create()` method is linked in some places in this package (where it makes sense IMHO) and linking to subclasses (creators) `create()` method doesn't make sense in context of `core` package. So I suppose it makes sense to have it here.

f1ames:
I have added `LoadedPlugins` typedef to fix return type in https://github.com/ckeditor/ckeditor5-core/pull/159/commits/a9c511e42e0a919cf9d50d18abd6091ed9760440.

src/editor/editor.js
pjasiun: > method is linked in some places in this package (where it makes sense IMHO)

I do not think so. There is no such method and there will be no such method. It is also not requested for all subclasses to create such a method, so I think you should update docs which liked to it.

src/plugin.js
```diff
@@ -169,3 +169,9 @@ mix( Plugin, ObservableMixin );
  * @method #destroy
  * @returns {null|Promise}
  */
+
+/**
+ * Array of loaded plugins.
```
ma2ciek, line 174: `An array`?

src/editor/editor.js
ma2ciek: And can't we have a `@returns {Promise.<module:core/editor/editor~Editor>} Promise that resolves to the editor instance once the editor is ready.` if we go with documenting this method? I don't get these multiple `@returns` tags honestly...

f1ames: It's hard to describe the initialization process when there is one step (`create()` method) which is missing from the base class but may be present (or at the current state always is) in the whole process. So from the point of view of the `Editor` class there is no `create()` method, but in real cases it is the core part / entry point of the whole process. For me it feels like something is not in the right place.

Anyway, I corrected the docs removing links to `create()` method trying to make them readable and easy to understand (see https://github.com/ckeditor/ckeditor5-core/pull/159/commits/b2b65f47d87d1a0f5c7194a34ac0c38917f9f3a9).

---

Btw. there are two places in the docs which I'm not sure how to handle:

1. https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/angular.html#editor-required - where should we link in such cases?
1. https://ckeditor.com/docs/ckeditor5/latest/framework/guides/architecture/core-editor-architecture.html#editor-classes - should `create()` just be skipped here as it is not present in the base class?

Reinmar: Let's continue this in https://github.com/ckeditor/ckeditor5/issues/1487.

src/editor/editor.js
```diff
@@ -49,8 +49,6 @@ export default class Editor {
 	/**
 	 * Creates a new instance of the Editor class.
 	 *
-	 * Usually, not to be used directly. See the static {@link module:core/editor/editor~Editor.create `create()`} method.
```
Reinmar: Revert change.

Reinmar: And all other changes which removed links to `create()`.

Reinmar:
I left some feedback in https://github.com/ckeditor/ckeditor5/issues/1487.

src/editor/editor.js
oleq: IMO we should either:
* not document `@param {Object} config` here because different editors can have different arguments. For instance the classic editor has `static create( sourceElementOrData, config )`. We mustn't enforce any order or type of arguments here – it's up to the developer what it looks like. 
* document it the way we implemented our 4 editors and hope this is not gonna change and will be enough for 99% of the cases, so 
```js
    /*
	 * @param {HTMLElement|String} sourceElementOrData The DOM element that will be the source for the created editor
	 * (on which the editor will be initialized) or initial data for the editor.
	 *
	 * If a source element is passed, then its contents will be automatically
	 * {@link module:editor-...Editor#setData loaded} to the editor on startup and the element
	 * itself will be used as the editor's editable element.
	 *
	 * If data is provided, then `editor.ui.view.editable.element` will be created automatically and needs to be added
	 * to the DOM manually.
	 * @param {module:core/editor/editorconfig~EditorConfig} config The editor configuration.
```

f1ames: 😭 The part you mentioned was actually a leftover of the old `create()` docs. I removed it and now we are left with docs proposed by @Reinmar in https://github.com/ckeditor/ckeditor5/issues/1487#issuecomment-458601552 - there are no `@param`s nor `@returns` which aligns with your first point:

> not document `@param {Object} config` here because different editors can have different arguments. For instance the classic editor has `static create( sourceElementOrData, config )`. We mustn't enforce any order or type of arguments here – it's up to the developer what it looks like.

PR #163, opened 2019-02-11
t/1517: Editor placeholder should be created from the "placeholder" attribute only when the source element is `<textarea>`.

oleq:

Docs: Editor placeholder should be created from the "placeholder" attribute only when the source element is `<textarea>`. (see ckeditor/ckeditor5#1517).

---

### Additional information

Part of https://github.com/ckeditor/ckeditor5/compare/t/1517?expand=1.

### DIFF
--- a/src/editor/editorconfig.jsdoc
+++ b/src/editor/editorconfig.jsdoc
@@ -187,11 +187,10 @@
  *			::before
  *		</p>
  *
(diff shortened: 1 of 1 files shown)


Reinmar:
I've got second thoughts. If it's named `data-placeholder` it's not going to work with textarea's `placeholder`. Do we need support for this attribute at all? Can't we handle this via the config only? And then, only for textareas, from an attribute.

PR #167, opened 2019-03-12
Reworded docs and renamed disable() and enable() to forceDisabled() and clearForceDisabled()

### LINKED ISSUE #165: Introduce `command#disable()` and `command#enable()`
More and more features are using "disable all/most commands" functionality (mostly in collaboration features though). It leads to duplicating a lot of code. Additionally, it is not really intuitive to disable a command through listening to `isEnabled`.

Hence, I propose to add this functionality to `Command` API. Of course, it will need to handle multiple disabling, so one algorithm will not re-enable lock set in another algorithm. This can be simply handled through passing identifiers.

```js
command.disable( 'myFeature' );
// ...
command.enable( 'myFeature' );
```

I think this is a better solution that returning enabling callback:

```js
const enable = command.disable();
// ...
enable();
```

Because of how we use disabling:

```js
// Disable all commands.
for ( const command of this.editor.commands.commands() ) {
	// Will become `command.disable( 'id' );`.
	command.on( 'set:isEnabled', forceDisable, { priority: 'highest' } );
	command.isEnabled = false;
}
// ...
for ( const command of this.editor.commands.commands() ) {
	// Will become `command.enable( 'id' );`.
	command.off( 'set:isEnabled', forceDisable );
	command.refresh();
}
```

If callback would be returned we would have to store all of them which doesn't make sense.

### LINKED ISSUE #2903: Introduce `command#disable()` and `command#enable()`
More and more features are using "disable all/most commands" functionality (mostly in collaboration features though). It leads to duplicating a lot of code. Additionally, it is not really intuitive to disable a command through listening to `isEnabled`.

Hence, I propose to add this functionality to `Command` API. Of course, it will need to handle multiple disabling, so one algorithm will not re-enable lock set in another algorithm. This can be simply handled through passing identifiers.

```js
command.disable( 'myFeature' );
// ...
command.enable( 'myFeature' );
```

I think this is a better solution that returning enabling callback:

```js
const enable = command.disable();
// ...
enable();
```

Because of how we use disabling:

```js
// Disable all commands.
for ( const command of this.editor.commands.commands() ) {
	// Will become `command.disable( 'id' );`.
	command.on( 'set:isEnabled', forceDisable, { priority: 'highest' } );
	command.isEnabled = false;
}
// ...
for ( const command of this.editor.commands.commands() ) {
	// Will become `command.enable( 'id' );`.
	command.off( 'set:isEnabled', forceDisable );
	command.refresh();
}
```

If callback would be returned we would have to store all of them which doesn't make sense.

scofalik:

Internal: Reworded docs and renamed `Command#disable()` and `Command#enable()` to `Command#forceDisabled()` and `Command#clearForceDisabled()`. Closes ckeditor/ckeditor5#2903.

### DIFF
--- a/src/command.js
+++ b/src/command.js
@@ -186,7 +186,7 @@ export default class Command {
 	/**
 	 * Clears forced disable previously set through {@link #clearForceDisabled}. See {@link #clearForceDisabled}.
 	 *
-	 * @param {String} id "Disable stack" identifier.
+	 * @param {String} id Unique identifier, equal to the one passed in {@link #forceDisabled} call.
 	 */
 	clearForceDisabled( id ) {
 		this._disableStack.delete( id );
--- a/src/command.js
+++ b/src/command.js
@@ -141,7 +141,7 @@ export default class Command {
 	/**
 	 * Disables the command.
 	 *
-	 * Command may become disabled by multiple features or algorithms (at once). When disabling a command, unique id should be passed
+	 * Command may be disabled by multiple features or algorithms (at once). When disabling a command, unique id should be passed
 	 * (e.g. feature name). The same identifier should be used when {@link #clearForceDisabled enabling back} the command.
 	 * The command becomes enabled only after all features {@link #clearForceDisabled enabled it back}.
 	 *
--- a/src/command.js
+++ b/src/command.js
@@ -120,9 +120,9 @@ export default class Command {
 		// By default commands are disabled when the editor is in read-only mode.
 		this.listenTo( editor, 'change:isReadOnly', ( evt, name, value ) => {
 			if ( value ) {
-				this.disable( 'readOnlyMode' );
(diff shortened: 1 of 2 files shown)


src/command.js
```diff
@@ -141,19 +141,40 @@ export default class Command {
 	/**
 	 * Disables the command.
 	 *
-	 * "Disable stack" is supported through identifiers. Command may be disabled by a multiple features/algorithms (at once).
-	 * When disabling a command, identifier is passed (and added to "disable stack"). The identifier is then used when
-	 * {@link #enable enabling back} the command. The command is actually enabled only after all features {@link #enable enabled it back}.
+	 * Command may become disabled by multiple features or algorithms (at once). When disabling a command, unique id should be passed
```
pjasiun:
```suggestion
	 * Command may be disabled by multiple features or algorithms (at once). When disabling a command, unique id should be passed
```

PR #173, opened 2019-05-07
Throw an error when user tries to use inappropriate editor over textarea

### LINKED ISSUE #1591: Confusing behavior of CKEditor 5 when launching the editor on a <textarea> element
There are several places where we show an example of Classic editor being launched on top of a `<textarea>` element:

 * https://cdn.ckeditor.com/#ckeditor5
 * https://ckeditor.com/docs/ckeditor5/latest/builds/guides/quick-start.html#classic-editor

If someone takes such an example and later tries to change the editor type (e.g. the Balloon editor), then will see that the editor no longer shows up:

```html
<!DOCTYPE html>
<html>
        <head>
                <meta charset="utf-8">
                <title>CKEditor</title>
                <script src="https://cdn.ckeditor.com/ckeditor5/12.0.0/balloon/ckeditor.js"></script>
        </head>
        <body>
                <textarea name="content" id="editor">This is some sample content.</textarea>
                <script>
                        BalloonEditor
                                .create( document.querySelector( '#editor' ) )
                                .then( editor => {
                                        console.log( 'hello' );
                                        console.log( editor );
                                        console.log( editor.getData() );
                                } )
                                .catch( error => {
                                        console.error( error );
                                } );
                </script>
        </body>
</html>
```

## Result in a browser:

screenshot 2019-03-04 at 17 48 49

## Expected result

1. <a href="https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-balloon_ballooneditor-BalloonEditor.html#static-function-create">BalloonEditor.create()</a> documentation could mention if there are any limitations on what can be passed as ` sourceElementOrData` argument.

2. The editor could log some meaningful error into the console when unsupported element is passed.

msamsel:

Feature: Provide method to prevent of creating editors in textarea. Closes ckeditor/ckeditor5#1591.

---

### Additional information
Related changes in other repositories:
* PR: ckeditor/ckeditor5-editor-inline#52 and [branch](https://github.com/ckeditor/ckeditor5-editor-inline/tree/t/ckeditor5/1591)
* PR: ckeditor/ckeditor5-editor-balloon#33 and [branch](https://github.com/ckeditor/ckeditor5-editor-balloon/tree/t/ckeditor5/1591)
* PR: ckeditor/ckeditor5-editor-decoupled#32 and [branch](https://github.com/ckeditor/ckeditor5-editor-decoupled/tree/t/ckeditor5/1591)


src/editor/editor.js
```diff
@@ -268,6 +270,39 @@ export default class Editor {
 		this.commands.execute( ...args );
 	}
 
+	/**
+	 * Methods creates rejected promise when provided parameter is a HTML Textarea element.
+	 * It resolves in any other case.
+	 *
+	 * @param {HTMLElement|String} sourceElementOrData The DOM element that will be the source for the created editor
+	 * or the editor's initial data.
+	 * @returns {Promise}
```
pomek: Why does this function return a promise if it isn't asynchronous itself? It should return `Boolean` or `undefined` and throws an error if specified arguments are invalid.

pomek: After checking related PR in balloon editor, I am for the second option (`undefined` + throw).

And the entire editor's code could be simplified:

```js
return new Promise( resolve => {	

    Editor.allowedElement( sourceElementOrData ); // Throw the error if specified parameter is invalid.
    // Editor.assertSourceElement() ?
    
    const editor = new this( sourceElementOrData, config );
    // ...
} );
```

msamsel: I use it in 3 editors (balloon, inline, decouple) to test what kind of element is used during editor creation. I felt that simpler is to kept error here rather than implementation of rejected promise in each of those editors (https://github.com/ckeditor/ckeditor5-editor-inline/blob/9b34b1106cc935394a67b4a07bf5a03ca00af5f1/src/inlineeditor.js#L192)
Is it an antipattern to use promise in cases like this?

pomek: It's more like an assertion check. You would like to check whether the specified parameter is proper from the editor point of view. The similar _mechanism_ we are using in Writer – https://github.com/ckeditor/ckeditor5-engine/blob/e0de1586620918f08854962e3ee323122c4e2059/src/model/writer.js#L1302-L1320.

That's why I am proposing changing the API and usage in `*Editor.create()`  classes.

> Is it an antipattern to use promise in cases like this?

In such case, we are forcing synchronous function to be asynchronous. For me, it sounds bad.

msamsel:
@pomek I've change solution to not include unnecessary promise. Thx for feedback.

pomek (approved):
Not sure whether we should stay with `_assertAllowedSourceElement` or change it to `_assertSourceElement` but the logic inside the method looks good.

src/editor/editor.js
```diff
@@ -268,6 +270,34 @@ export default class Editor {
...
 	}
 
+	/**
+	 * Throws `editor-wrong-element` when checked element is `<textarea>`.
+	 *
+	 * @protected
+	 * @param {HTMLElement|String} sourceElementOrData The DOM element that will be the source for the created editor
+	 * or the editor's initial data.
+	 */
+	static _assertAllowedSourceElement( sourceElementOrData ) {
```
Reinmar: You added this method to the base editor. But there are editor types where textarea is a valid source element. So, in other words, this method is not that generic and should either be made more specific or moved out of here.

The other problem with it is that it does not actually check allowed elements yet. It checks only the textarea case. So not only this method is not defined in the right place, but also it's called incorrectly.

I think the easiest solution here would be to have this duplicated – there's very little code here. The only thing which needs to be DRY-ed is the error message, so we could keep it here and document in the places where we'll be firing it that it's defined here.

src/editor/editor.js
```diff
@@ -268,6 +270,34 @@ export default class Editor {
...
+		if ( isElement( sourceElementOrData ) && sourceElementOrData.tagName.toLowerCase() === 'textarea' ) {
```
Reinmar: Why `toLowerCase()`?

msamsel: You're right, it's completely unnecessary here.

src/editor/editor.js
msamsel: I've move element check directly to `create` methods of each class. It's going to be similar to checks for initial data:
https://github.com/ckeditor/ckeditor5-editor-balloon/blob/d5c4162fe018c65604a59c5cd0d0b0f5e53c6574/src/ballooneditor.js#L209-L212

msamsel:
@jodator I asked you for review, as @Reinmar won't have time in near time to make another review of this PR.
Please notice that main changes are located in related repositories and PRs.

jodator:
I've extracted this one: https://github.com/ckeditor/ckeditor5/issues/1806 to review docs later on.

jodator:
@msamsel I found that all that code is practically duplicated over those 3 repositories. I think that it would make sense to create a single assertion method for that in core & import it in proper editors. Something similarly to `attachToForm()` method from `core/src/editor/utils`.

Maybe `assertNotUsedOverTextareaElement( initialDataOrElement )`? But if you have better name please propose something :D Also move the error docs there.

It also makes sense to extract this if we ever add another editor type.

jodator (changes requested):
@msamsel I found that all that code is practically duplicated over those 3 repositories. I think that it would make sense to create a single assertion method for that in core & import it in proper editors. Something similarly to `attachToForm()` method from `core/src/editor/utils`.

Maybe `assertNotUsedOverTextareaElement( initialDataOrElement )`? But if you have better name please propose something :D Also move the error docs there.

It also makes sense to extract this if we ever add another editor type.

msamsel:
> @msamsel I found that all that code is practically duplicated over those 3 repositories. I think that it would make sense to create a single assertion method for that in core & import it in proper editors. Something similarly to attachToForm() method from core/src/editor/utils.

@jodator that was my thought with that, maybe not in proper place, but to keep common logic in editor's code. It was even implemented here:
https://github.com/ckeditor/ckeditor5-core/pull/173/commits/b80e923e72a741ef7c95e94bad9249ed08f2e636#diff-ccd92a4dde1c850ebd60f90f39638203L280
However @Reinmar suggest to duplicate this code in specific editor implementations:
https://github.com/ckeditor/ckeditor5-core/pull/173#discussion_r286366282

> It also makes sense to extract this if we ever add another editor type.

AFAIR I talk about this case with @Reinmar and There is very little chance that we introduce new editor type any time soon.

jodator (approved):
Fine enough - the place wasn't good. But we can leave those check as they are.

PR #177, opened 2019-06-06
Re-enable memory tests.

jodator:

Internal: Re-enable memory tests.

---

### Additional information

* It looks like the newest Chrome 75.x does not longer break our memory tests.


Reinmar:
OK, let's see how it works on our CI :D

PR #179, opened 2019-06-14
T/ckeditor5/1636: MultiCommand

jodator:

Feature: Introduce `MultiCommand` that allows to group many commands as one.

---

### Additional information

This is part of [IndentBlock PR](https://github.com/ckeditor/ckeditor5/pull/1816).

The changes in core:
1. Created Indent feature plugins:
    - `Indent` - typical glue plugin.
    - `IndentEditing` - introduces the "multi" `'indent'` and `'outdent'` commands.
    - `IndentUI` - introduces the `'indent'` and `'outdent'` toolbar buttons.

    We talked about defining `Indent` feature (buttons) with @Reinmar but I'm divided on where to  put this feature. From one side this feature might not be needed by some - thus it might be better to put it in... :drum: ...another repository as a separate `ckeditor5-indent` feature. But OTOH we will probably include this in most of the builds to use `'indent'` and `'outdent'` buttons.

2. The `MultiCommand` that allows to register many commands under one alias.

    I've created the `MultiCommand` as a way to programmatically register other indentation commands under one command so the UI buttons would attach to one command only event thought there are [voices against](https://github.com/ckeditor/ckeditor5/issues/1636#issuecomment-499860457) (by @pjasiun dunno @Reinmar voice on this).

    The command allows other features to register a child command (if it is defined in command collection). The `MultiCommand` changes its `isEnabled` state to `true` if any of registered commands is enabled. It also executes first enabled command (if any).

    I've created this command as we want to expose `'indent'` and `'outdent'` buttons that will enable to indent list and indent blocks. We already have `'indentList'` (I'll skip the `outdent` as it is similar) command and a newly introduced `'indentBlock'` command. We want to have one button for both (as in CKEditor 4). Now, I don't like the idea of overriding, ie `indentBlock` command by `indentList` command if the command is in list and not in a block. This isn't clean and might introduce some WTFs to others. Now the `MultiCommand` might be just `InputCommand` with the same interface as we have only one usage for such construct now and in forseable future. But I'm strongly against any `if`s or other hacks in one commands that would tightly couple both features. ([early stage POC](https://github.com/ckeditor/ckeditor5/issues/1636#issuecomment-499814586) with buttons)

    The `MultiCommand` allows to  have any of indent feature to be present and does not require loading of other.


jodator:
I've invited you @pjasiun to review this alongisde @oleq just because the previous comments on this issue.

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+		command.execute( args );
+	}
+
+	/**
+	 * Registers a command as child command.
+	 *
+	 * @param {module:core/command~Command} command
+	 */
+	registerChildCommand( command ) {
+		this._childCommands.add( { command } );
```
Reinmar: Why do you wrap it with an object?

Reinmar: Because we use `Collection`? Then why using `Collection`?

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+
+		command.execute( args );
+	}
+
+	/**
+	 * Registers a command as child command.
+	 *
+	 * @param {module:core/command~Command} command
+	 */
+	registerChildCommand( command ) {
```
Reinmar, line 72: Just `addChildCommand()`?

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+ *		const multiCommand = new MultiCommand( editor );
+ *
+ *		const commandFoo = new Command( editor );
+ *		const commandBar = new Command( editor );
+ *
+ *		// Register child commands.
+ *		multiCommand.registerChildCommand( commandFoo );
+ *		multiCommand.registerChildCommand( commandBar );
+ *
+ *		// Enable one of commands
```
Reinmar:
```suggestion
 *		// Enable one of the commands.
```

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+import Collection from '@ckeditor/ckeditor5-utils/src/collection';
+
+/**
+ * @module core/multicommand
+ */
+
+/**
+ * A CKEditor command that aggregates other commands.
+ *
+ * This command is used to proxy multiple commands. The multi command is enabled when one of its registered child commands is enabled.
```
Reinmar:
```suggestion
 * This command is used to proxy multiple commands. The multi-command is enabled when at least one of its registered child commands is enabled.
```

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+ * Whe executing multi command the first command that is enabled will be executed.
```
Reinmar:
```suggestion
 * When executing a multi-command the first command that is enabled will be executed.
```

src/multicommand.js
Reinmar: Although, "register" reads quite well in the docs. So I don't have a strong opinion on that.

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+	 * Executes the first of it registered child commands.
+	 */
+	execute( ...args ) {
+		const { command } = this._getFirstEnabledCommand();
+
+		command.execute( args );
+	}
+
+	/**
+	 * Registers a command as child command.
```
Reinmar:
```suggestion
	 * Registers a child command.
```

src/indent/indent.js
```diff
@@ -0,0 +1,50 @@
...
+
+import Plugin from '../plugin';
+
+import IndentEditing from './indentediting';
+import IndentUI from './indentui';
+
+/**
+ * The indent feature.
+ *
+ * This plugin acts as a single entry point plugin for other features that implements indenting of elements like list or paragraphs.
```
Reinmar:
```suggestion
 * This plugin acts as a single entry point plugin for other features that implement indenting of elements like lists or paragraphs.
```

src/indent/indent.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * The compatible features are:
```
Reinmar:
```suggestion
 * The compatible features are:
 
```

src/indent/indent.js
```diff
@@ -0,0 +1,50 @@
...
+ * - the {@link module:list/list~List} or {@link module:list/listediting~ListEditing} feature for list indentation
+ * - the {@link module:indent-block/indentblock~IndentBlock} feature for indenting text blocks like paragraphs or headings
```
Reinmar:
```suggestion
 * * the {@link module:list/list~List} or {@link module:list/listediting~ListEditing} feature for list indentation
```

src/indent/indent.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This is a "glue" plugin which loads the following plugins:
+ *
+ * * The {@link module:core/indent/indentediting~IndentEditing indent editing feature} and
+ * * The {@link module:core/indent/indentui~IndentUI indent UI feature}.
+ *
+ * The dependent plugins registers the `'indent'` and `'outdent'` commands and it introduces the `'indent'` and `'outdent'` buttons
```
Reinmar:
```suggestion
 * The dependent plugins register the `'indent'` and `'outdent'` commands and it introduce the `'indent'` and `'outdent'` buttons
```

src/indent/indent.js
```diff
@@ -0,0 +1,50 @@
...
+ * that allow to increase or decrease text indentation of supported elements.
```
Reinmar:
```suggestion
 * which allow to increase or decrease text indentation of supported elements.
```

src/multicommand.js
jodator: The "register" implies something more than just adding it (like adding to a collection). So if it also reads well in the docs I'll leave it as is.

src/multicommand.js
jodator: I think that I've started with collection to use some of its delegation powers but it turned out I use as an array. I'll change that.

src/multicommand.js
Reinmar: Good point. We actually start watching those commands.

src/multicommand.js
```diff
@@ -0,0 +1,99 @@
...
+		 * @private
+		 */
+		this._childCommands = [];
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	refresh() {
+		// Override base command refresh(): the command's state is changed when one of child commands changes states.
```
Reinmar, line 55: I became unsure about this. I'll report a followup, though.

PR #186, opened 2019-07-10
Removed usage of logger.

ma2ciek:

Internal: Removed usage of the CKEditor 5 logger. Part of ckeditor/ckeditor5#383.

---

### Additional information

### DIFF
--- a/src/plugincollection.js
+++ b/src/plugincollection.js
@@ -7,8 +7,9 @@
  * @module core/plugincollection
  */
 
-import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
-import log from '@ckeditor/ckeditor5-utils/src/log';
+/* globals console */
+
+import CKEditorError, { attachLinkToDocumentation } from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
 import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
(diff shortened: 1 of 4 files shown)


src/plugincollection.js
```diff
@@ -189,7 +190,7 @@ export default class PluginCollection {
 			const errorMsg = 'plugincollection-plugin-not-found: Some plugins are not available and could not be loaded.';
 
 			// Log the error so it's more visible on the console. Hopefully, for better DX.
-			log.error( errorMsg, { plugins: missingPlugins } );
+			console.error( attachLinkToDocumentation( errorMsg ), { plugins: missingPlugins } );
```
pjasiun, line 193: This should be `CKEditorError` exception.

src/plugincollection.js
```diff
@@ -231,7 +232,9 @@ export default class PluginCollection {
 					 * @error plugincollection-load
 					 * @param {String} plugin The name of the plugin that could not be loaded.
 					 */
-					log.error( 'plugincollection-load: It was not possible to load the plugin.', { plugin: PluginConstructor } );
+					console.error( attachLinkToDocumentation(
```
pjasiun, line 235: This should be `CKEditorError` exception.

src/plugincollection.js
```diff
@@ -371,8 +374,8 @@ export default class PluginCollection {
 			 * @param {Function} plugin1 The first plugin constructor.
 			 * @param {Function} plugin2 The second plugin constructor.
 			 */
-			log.warn(
-				'plugincollection-plugin-name-conflict: Two plugins with the same name were loaded.',
+			console.warn(
```
pjasiun: This should be `CKEditorError` exception.

src/plugincollection.js
ma2ciek: We throw an error below.

src/plugincollection.js
ma2ciek: No. We throw the correct error below.

src/plugincollection.js
pjasiun: :+1:

PR #196, opened 2019-09-16
t/ckeditor5/416d: Created a "three vertical dots" icon

oleq:

Internal: Created a "three vertical dots" icon for the toolbar grouping feature (see ckeditor/ckeditor5#416). 

---

### Additional information

Part of https://github.com/ckeditor/ckeditor5-ui/pull/523.


theme/icons/three-vertical-dots.svg
```diff
@@ -0,0 +1 @@
+<svg width="10" height="20" xmlns="http://www.w3.org/2000/svg"><g><circle cx="9.5" cy="4.5" r="1.5"/><circle cx="9.5" cy="10.5" r="1.5"/><circle cx="9.5" cy="16.5" r="1.5"/></g></svg>
```
dkonopka: We are not using `width`/`height` property for icons, let's change it to the `viewBox` (with 0 0 20 20).

 Also, I would take into consideration to transform svg content to `<path>` instead a group of `<circle>`:

oleq:
```suggestion
<svg viewBox="0 0 10 20" xmlns="http://www.w3.org/2000/svg"><circle cx="9.5" cy="4.5" r="1.5"/><circle cx="9.5" cy="10.5" r="1.5"/><circle cx="9.5" cy="16.5" r="1.5"/></svg>
```

oleq: Is path any better than circle @dkonopka?

dkonopka: I was thinking about the smaller weight of `.svg` file, but I'm not so sure.

PR #200, opened 2019-10-23
Extend `Editor` class with debug utils

ma2ciek:

Internal: Extended `Editor` class with debug tools. Part of ckeditor/ckeditor5#5649, https://github.com/ckeditor/ckeditor5-engine/pull/1805.

---

### Additional information


src/editor/editor.js
```diff
@@ -293,6 +309,23 @@ export default class Editor {
...
+	// @if CK_DEBUG_ENGINE // logView( version ) {
+	// @if CK_DEBUG_ENGINE //	this.editing.view.document.log( version );
+	// @if CK_DEBUG_ENGINE // }
+
+	// @if CK_DEBUG_ENGINE // logDocuments( version ) {
+	// @if CK_DEBUG_ENGINE //	version = version === null ? this.model.document.version : version;
+
+	// @if CK_DEBUG_ENGINE //	this.logModel( version );
+	// @if CK_DEBUG_ENGINE //	this.logView( version );
+	// @if CK_DEBUG_ENGINE // }
```
pjasiun, line 328: I think it would be great to keep all the code related to engine debugging in the `ckeditor5-engine` module. Not only because it is called `CK_DEBUG_ENGINE`, but also because the main editor class is kind of the main editor API. Many people will read this code to learn how the editor is created or to debug some problems (not necessarily related to the engine). I think that we should not add such code here since it makes reading this main file harder.

pjasiun: I think that `this.model.document.log( version );` and `this.editing.view.document.log( version );` is enough. We do not need these additional methods on the editor. The rest of the code I believe can be moved to the engine.

PR #201, opened 2019-11-18
i/5692: Added documentation for the config.toolbar.shouldNotGroupWhenFull option

### LINKED ISSUE #5692: Add possibility to toggle toolbar items grouping in a build
Is there an easy way to implement a multi-row toolbar like in previous ckeditor versions?  I like the dynamic grouping, but I am also in need of a multi-row toolbar.  

I did not find anything describing this by looking at the documentation or searching the issues, thanks!

oleq:

Docs: Added documentation for the `config.toolbar.shouldGroupWhenFull` option. Closes ckeditor/ckeditor5#5692.

---

### Additional information

Requires:
* https://github.com/ckeditor/ckeditor5-editor-classic/pull/100
* https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/45

### DIFF
--- a/src/editor/editorconfig.jsdoc
+++ b/src/editor/editorconfig.jsdoc
@@ -144,7 +144,9 @@
  *			toolbar: {
  *				items: [ 'bold', 'italic', '|', 'undo', 'redo' ],
  *
- *				viewportTopOffset: 30
+ *				viewportTopOffset: 30,
+ *
+ *				shouldGroupWhenFull: true
  *			}
  *		};
  *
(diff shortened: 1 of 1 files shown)


src/editor/editorconfig.jsdoc
```diff
@@ -162,6 +164,11 @@
  * * **`toolbar.viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar.
  * Useful when a page with which the editor is being integrated has some other sticky or fixed elements
  * (e.g. the top menu). Thanks to setting the toolbar offset the toolbar will not be positioned underneath or above the page's UI.
+ * * **`toolbar.shouldGroupWhenFull`** &ndash; When `true` (default), the toolbar will group its items that
```
oleq: This documentation entry is growing (`items`, `viewportTopOffset` and now `shouldGroupWhenFull`). I know it's convenient for developers to keep it in a single place (`module:core/editor/editorconfig~EditorConfig#toolbar`) but if we get more options like this, I think it's time to define some `@interface ToolbarConfig` and put the detailed documentation there.

src/editor/editorconfig.jsdoc
```diff
@@ -144,7 +144,9 @@
  *			toolbar: {
  *				items: [ 'bold', 'italic', '|', 'undo', 'redo' ],
  *
- *				viewportTopOffset: 30
+ *				viewportTopOffset: 30,
+ *
+ *				shouldGroupWhenFull: true
```
Reinmar:
```suggestion
 *				shouldGroupWhenFull: false
```

No one will be setting it to `true` as this is the default.

src/editor/editorconfig.jsdoc
Reinmar: As long as those options are fairly simple (boolean, number, simple array), I'd keep them locally here because, as you mentioned, that more convenient to read. But I agree that at some point we may need to extract it. But then we can keep a lot of information here locally, and have details in the interface.

src/editor/editorconfig.jsdoc
Reinmar: Actually, I'd rather rename it to `shouldNotGroupWhenFull` to avoid a flag which is default by true. Then you'll also have less issues with handling the default in the other two PRs.

Reinmar (changes requested):
As commented.

PR #204, opened 2019-12-02
Plugins outside of the editor

### LINKED ISSUE #5891: Plugins outside of the editor - editor context
The time has come when some of our plugins evolved so much that they can be used outside of the editor. Such plugins have to stay compatible with the editor (like it was before) but also have to be ready to be used without the editor, or even with multiple editors at once.

After some researches and discussions, we decided to introduce a `Context` class, that's gonna be an environment for the plugins used outside of the editor. Also, a new type of plugin, a `ContextPlugin` will be introduced to represents plugins compatible with the context and the editor. The current `Plugin` class will be compatible only with the editor like it was before.

Context API will be similar to the editor API but will be limited only to some properties:

```javascript
Context.plugins
Context.config
Context.locale
```

# Scenarios

## Single editor usage

For the single editor, not much change. Editor, when created, initialize additional property `editor.context`, with an empty context used only by this editor. There are no changes in the editor API for basic usage.

## Shared editor usage

Assume that the `Comments` plugin requires the `Sidebar` plugin.

```js
const context = async Context.create( {
	plugins: [ Sidebar ],
	sidebar: {
		container: sidebarElement
	}
} );

const editor1 = async ClassicEditor.create( editorElement1, {
	context: context,
	plugins: [ ArticlePlugins, Comments ]
} );

const editor2 = async ClassicEditor.create( editorElement2, {
	context: context,
	plugins: [ ArticlePlugins, Comments ]
} );
```

If the `context` property is passed to the editor it means that:
 - `context.plugins` (instances) should be added to the editor plugins,
 - `editor.config` should be extended by the `context.config`,
 - `editor.locale` should use `context.locale`.

Note, that this way, both editors will share the same `Sidebar` plugin. When any of these editors use `editor.plugins.get( 'Sidebar' )` they will get the same instance.

When it comes to dependencies:
 - it should be possible, for the context plugins to requires another context plugins,
 - it should be possible, for the editor plugins to requires context plugins,
 - it should NOT be possible, for the context plugins to require editor plugins.

oskarwrobel:

Feature: Introduced `Context` class as an environment for a new `ContextPlugin` class that represents features that work out of the editor. Closes ckeditor/ckeditor5#5891.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-utils/pull/316.


src/context.js
```diff
@@ -0,0 +1,136 @@
...
+ * @module core/context
+ */
+
+import Config from '@ckeditor/ckeditor5-utils/src/config';
+import PluginCollection from './plugincollection';
+import Locale from '@ckeditor/ckeditor5-utils/src/locale';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+/**
+ * @TODO
```
oskarwrobel: @Reinmar, could you help with that?

src/context.js
```diff
@@ -0,0 +1,162 @@
...
+	 * @returns {Promise.<module:core/plugin~LoadedPlugins>} A promise which resolves
+	 * once the initialization is completed providing an array of loaded plugins.
+	 */
+	initPlugins() {
+		const plugins = this.config.get( 'plugins' ) || [];
+
+		for ( const Plugin of plugins ) {
+			if ( typeof Plugin != 'function' ) {
+				throw new CKEditorError(
+					'context-initplugins-constructor-only: Only constructor is allowed as a Context plugin.',
```
scofalik, line 123: Missing error documentation.

src/context.js
```diff
@@ -0,0 +1,162 @@
...
+					null,
+					{ Plugin }
+				);
+			}
+
+			if ( Plugin.isContextPlugin !== true ) {
+				throw new CKEditorError(
```
scofalik, line 135: Missing error documentation.

src/context.js
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		return this.plugins.init( plugins );
+	}
+
+	/**
+	 * Returns context configuration which will be copied to editors created using this context.
+	 *
+	 * The configuration returned by this method has removed plugins configuration - plugins are shared with all editors
+	 * in a special way.
```
scofalik: "in a special way" => "through another mechanism"

src/editor/editor.js
```diff
@@ -74,7 +89,21 @@ export default class Editor {
...
 		 * @member {module:core/plugincollection~PluginCollection}
 		 */
-		this.plugins = new PluginCollection( this, availablePlugins );
+		this.plugins = new PluginCollection( this, availablePlugins, this.context.plugins );
+
+		/**
+		 * @readonly
+		 * @type {module:utils/locale~Locale}
+		 */
+		this.locale = this.context.locale;
```
scofalik: It doesn't look like it is correct to take locale always from the context, although this might be something that we (you and @pjasiun) agreed on and I am not aware.

But anyway, does it work if there's no context specified but the language configuration is specified in the editor? I can't see how it is working 🤔 .

src/editor/editor.js
```diff
@@ -48,9 +48,24 @@ export default class Editor {
...
+		this.context = config.context || new Context();
+		this.context.addEditor( this );
+
+		/**
+		 * `true` when context is created by this editor or `false` when context is injected to the editor.
+		 *
+		 * @private
+		 * @type {Boolean}
+		 */
+		this._isContextHost = !config.context;
```
scofalik: I don't like that name. `_didCreateContext` or `_hasOwnContext` or `_wasContextPassed` or `_managesOwnContext` ... there are a lot of names. When I saw this one, I thought it has an opposing meaning :). I mean, it doesn't tell you much cause you need to explain what "host" means in this ... context (lol).

src/editor/editor.js
```diff
@@ -247,6 +257,12 @@ export default class Editor {
 
 		return readyPromise
 			.then( () => {
+				this.context.removeEditor( this );
+
+				if ( this._isContextHost ) {
```
scofalik: I wonder ... If this created it's own context, do we actually need to destroy it?

1. All the editors in the context will get destroyed anyway.
2. The context should not have any plugins initialized.

It's probably still safer to destroy the context anyway, in case of some future changes in `Context#destroy()`. But I don't like that we need this property just for this check :P. (We could check something like `context._editors.length` but that might fail for contexts which were created outside `Editor` -- "real contexts").

src/plugincollection.js
```diff
@@ -23,37 +23,56 @@ export default class PluginCollection {
 	/**
 	 * Creates an instance of the PluginCollection class.
 	 * Allows loading and initializing plugins and their dependencies.
+	 * Allows to provide a list of already loaded plugins, these plugins won't be destroyed along with this collection.
 	 *
 	 * @param {module:core/editor/editor~Editor} editor
```
scofalik: Editor or Context.

src/plugincollection.js
```diff
@@ -23,37 +23,56 @@ export default class PluginCollection {
...
 	 * used in `config.plugins` or `config.removePlugins` by names.
+	 * @param {Iterable.<Array>} externalPlugins List of already initialized plugins represented by a
+	 * `[ PluginConstructor, pluginInstance ]` pair.
 	 */
-	constructor( editor, availablePlugins = [] ) {
+	constructor( editor, availablePlugins = [], externalPlugins = [] ) {
 		/**
 		 * @protected
-		 * @member {module:core/editor/editor~Editor} module:core/plugin~PluginCollection#_editor
+		 * @type {module:core/editor/editor~Editor}
```
scofalik: Editor or Context.

src/plugincollection.js
```diff
@@ -319,10 +362,13 @@ export default class PluginCollection {
 	 * @returns {Promise}
 	 */
 	destroy() {
-		const promises = Array.from( this )
-			.map( ( [ , pluginInstance ] ) => pluginInstance )
-			.filter( pluginInstance => typeof pluginInstance.destroy == 'function' )
-			.map( pluginInstance => pluginInstance.destroy() );
+		const promises = [];
```
scofalik, line 365: Not sure if this refac was needed 🤷‍♂ . Didn't you want to just move function from `.filter()` to external function and expand it with the new condition?

src/editor/editor.js
```diff
@@ -48,9 +48,24 @@ export default class Editor {
...
-	 * @param {Object} [config] The editor config.
+	 * @param {Object} [config={}] The editor config.
 	 */
-	constructor( config ) {
+	constructor( config = {} ) {
+		/**
+		 * @readonly
+		 * @type {module:core/context~Context}
+		 */
+		this.context = config.context || new Context();
```
scofalik: Do we need that to be public? 🤔

scofalik (changes requested):
I miss changes in `plugin.js`. I'd `add isContextPlugin()` there and also I think that `PluginInterface` might need some changes too.

Also, I am not sure, whether `Context` and `Editor` should share some kind of interface too. It would be easier to use that interface as a type in JSDoc and in docs in general. Maybe even a mixin or a class if we could be sure that some code could be shared between them. Any comment on that matter?

src/editor/editor.js
oskarwrobel: I had exactly the same thoughts. That's why I created an additional property.

src/editor/editor.js
oskarwrobel: For me `isContextHost` is a quite clear name, but names proposed by you sound good as well.

src/editor/editor.js
oskarwrobel: You are right. In this case, editor language configuration won't be used. It's a bug.

src/editor/editor.js
oskarwrobel: 🤔 looks like it doesn't have to be public.

src/plugincollection.js
oskarwrobel: 🤷‍♂

oskarwrobel:
> Also, I am not sure, whether Context and Editor should share some kind of interface too. It would be easier to use that interface as a type in JSDoc and in docs in general. Maybe even a mixin or a class if we could be sure that some code could be shared between them. Any comment on that matter?

It's an interesting idea, but I think we should create a follow-up for that. Some docs especially `Plugin` and `PluginCollection` should be verified according to `Context`. Context class needs some general docs as well. I will create a followup about improving docs.

src/context.js
oskarwrobel: Done.

src/context.js
oskarwrobel: Done.

src/context.js
oskarwrobel: Done.

src/editor/editor.js
oskarwrobel: I fixed that.

src/editor/editor.js
oskarwrobel: I moved this property from the editor to the context (and changed the name 😄).

src/editor/editor.js
oskarwrobel: It is protected now.

src/context.js
```diff
@@ -0,0 +1,201 @@
...
+		this.t = this.locale.t;
+
+		/**
+		 * When the context is created by an editor then the editor instance is
+		 * stored as an owner of this context.
+		 *
+		 * @readonly
+		 * @type {Boolean}
+		 */
+		this.isCreatedByEditor = false;
```
scofalik:
```suggestion
		this.wasCreatedByEditor = false;
```

scofalik (approved):

scofalik:
This is fine for me after the changes. But we need to make a decision about docs. I or Oskar could write some initial docs and make an issue for @Reinmar to do it correctly (and probably add/enhance some of the non-API docs).

Also, I'd like to see a follow up for "editor interface".

scofalik:
Okay, I filled missing docs and we have a follow-up to review them. So, @Reinmar, if you are fine with those changes, feel free to merge them.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+ * or/and plugins that work outside of an editor. Use it instead of {@link module:core/editor/editor~Editor.create `Editor.create()`}
+ * in advanced application integrations.
+ *
+ * All configuration options passed to a `Context` will be used as default options for editor instances initialized in that context.
+ *
+ * {@link module:core/contextplugin~ContextPlugin `ContextPlugin`s} passed to a `Context` instance will be shared among all
+ * editor instances initialized in that context. These will be the same plugin instances for all the editors.
+ *
+ * **Note:** `Context` can only be initialized with {@link module:core/contextplugin~ContextPlugin `ContextPlugin`s}
+ * (e.g. {@glink features/collaboration/comments/comments comments}). Regular {@link module:core/plugin~Plugin `Plugin`s} require an
```
Reinmar: I think this will throw on CI and for us locally. AFAIR those links to collab features have to be created with standard Markdown syntax because we often build the docs of just one project.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+ * editor instance to work and cannot be added to a `Context`.
+ *
+ * **Note:** You can add `ContextPlugin` to an editor instance, though.
+ *
+ * If you are using multiple editor instances on one page and use any `ContextPlugin`s, create `Context` to share configuration and plugins
+ * among those editors. Some plugins will use the information about all existing editors to better integrate between them.
+ *
+ * If you are using plugins that do not require an editor to work (e.g. {@glink features/collaboration/comments/comments comments})
```
Reinmar: Same here.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+		 */
+		this.t = this.locale.t;
+
+		/**
+		 * `true` when the context is created by an editor `false` otherwise.
+		 *
+		 * @readonly
+		 * @type {Boolean}
+		 */
+		this.wasCreatedByEditor = false;
```
Reinmar: My first question was supposed to be – do we need this property to be public? It looks a bit fishy.

Then I checked that it's used in just one place (I've got a good smell :D) to check whether the context should be destroyed together with an editor. The question for that part is – should the editor manage the context or vice versa? Perhaps the editor should just tell the context that it's being destroyed and then the context may make a decision itself what should happen with it.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+
+		/**
+		 * List of editors to which this context instance is injected.
+		 *
+		 * @private
+		 * @type {Set<module:core/editor/editor~Editor>}
```
Reinmar:
```suggestion
		 * @type {Set.<module:core/editor/editor~Editor>}
```

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+		/**
+		 * `true` when the context is created by an editor `false` otherwise.
+		 *
+		 * @readonly
+		 * @type {Boolean}
+		 */
+		this.wasCreatedByEditor = false;
+
+		/**
+		 * List of editors to which this context instance is injected.
```
Reinmar: Do we inject the context to the editor or add editors to the context? The API of this class says the latter. It'd be good to solve this relation because it touches the previous comment – who should take care of whom?

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+
+	/**
+	 * Adds a reference to the editor which is used with this context.
+	 *
+	 * When the context is created by the editor it is additionally
+	 * marked as a {@link ~Context#wasCreatedByEditor} what is used
+	 * in the destroy chain.
+	 *
+	 * @param {module:core/editor/editor~Editor} editor
+	 * @param {Boolean} isContextOwner
```
Reinmar:
```suggestion
	 * @param {Boolean} isContextOwner Mark this context as created by the given editor.
```

src/context.js
Reinmar: Why do we need this check? Does it have something to do with the fact that to a standard editor you can pass strings (if builtin plugins were provided)?

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+	 * @returns {Promise.<module:core/plugin~LoadedPlugins>} A promise which resolves
+	 * once the initialization is completed providing an array of loaded plugins.
+	 */
+	initPlugins() {
+		const plugins = this.config.get( 'plugins' ) || [];
+
+		for ( const Plugin of plugins ) {
+			if ( typeof Plugin != 'function' ) {
+				/**
+				 * Only constructor is allowed as a Context plugin.
```
Reinmar: Make this a link to the `ContextPlugin` class. The goal of those messages here is that we can extend the raw error message. On the console, we cannot link to things that easily like here.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+				throw new CKEditorError(
+					'context-initplugins-constructor-only: Only constructor is allowed as a Context plugin.',
+					null,
+					{ Plugin }
+				);
+			}
+
+			if ( Plugin.isContextPlugin !== true ) {
+				/**
+				 * Only plugin marked as a ContextPlugin is allowed to be used with a context.
```
Reinmar: Link to `ContextPlugin` class.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+			if ( ![ 'plugins', 'removePlugins', 'extraPlugins' ].includes( name ) ) {
+				result[ name ] = this.config.get( name );
+			}
+		}
+
+		return result;
+	}
+
+	/**
+	 * Destroys the context instance, releasing all resources used by it.
```
Reinmar: Should mention that it destroys the editors too.

src/context.js
```diff
@@ -0,0 +1,261 @@
...
+			.then( () => this.plugins.destroy() );
+	}
+
+	/**
+	 * Creates and initializes a new context instance.
+	 *
+	 *		const commonConfig = { ... }; // Configuration for all the plugins and editors.
+	 *		const editorPlugins = [ ... ]; // Regular `Plugin`s here.
+	 *
+	 *		const context = await Context.create( {
```
Reinmar: Please avoid awaits in the docs since we use promise API in all other places.

src/context.js
oskarwrobel: > The question for that part is – should the editor manage the context or vice versa? Perhaps the editor should just tell the context that it's being destroyed and then the context may make a decision itself what should happen with it.

It depends on who has created the context. If the context was created outside of the editor, the editor should not manage it. But, if the context was created by the editor (single editor creates an internal context to simplify some code) then the context should be destroyed inside the editor destroy chain.

src/contextplugin.js
```diff
@@ -0,0 +1,50 @@
...
+	 * Creates a new plugin instance.
+	 *
+	 * @param {module:core/context~Context|module:core/editor/editor~Editor} context
+	 */
+	constructor( context ) {
+		/**
+		 * The context instance.
+		 *
+		 * @readonly
+		 * @type {module:core/context~Context|module:core/editor/editor~Editor}
```
Reinmar, line 41: WDYT about moving the `Context` class to the `editor/` directory? It's strongly editor-related.

src/contextplugin.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @module core/contextplugin
+ */
+
+import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import mix from '@ckeditor/ckeditor5-utils/src/mix';
+
+/**
+ * The base class for {@link module:core/context~Context} plugin classes.
```
Reinmar, line 14: > A context plugin can either be initialized for an editor or for a context. In other words, it can either work within one editor instance or with one or more editor instances that use a single context. It's the context plugin's role to implement handling for both modes.

Is the above correct? Cause I'm not sure how it's supposed to work. If I'm correct, please add it to the docs. If not, explain me how it's supposed to work :D

oskarwrobel: It's correct 👍

src/context.js
oskarwrobel: >  Does it have something to do with the fact that to a standard editor you can pass strings (if builtin plugins were provided)?

Yep.

src/plugincollection.js
```diff
@@ -23,37 +23,56 @@ export default class PluginCollection {
...
+	 * @param {module:core/editor/editor~Editor|module:core/context~Context} context
 	 * @param {Array.<Function>} [availablePlugins] Plugins (constructors) which the collection will be able to use
 	 * when {@link module:core/plugincollection~PluginCollection#init} is used with plugin names (strings, instead of constructors).
 	 * Usually, the editor will pass its built-in plugins to the collection so they can later be
 	 * used in `config.plugins` or `config.removePlugins` by names.
+	 * @param {Iterable.<Array>} externalPlugins List of already initialized plugins represented by a
+	 * `[ PluginConstructor, pluginInstance ]` pair.
 	 */
-	constructor( editor, availablePlugins = [] ) {
+	constructor( context, availablePlugins = [], externalPlugins = [] ) {
```
Reinmar: Plugin instances. Plugins are the constructors. Here, we're talking about instances. Also, this could be a map perhaps?

src/editor/editor.js
```diff
@@ -63,8 +73,8 @@ export default class Editor {
 		 * @member {module:utils/config~Config}
 		 */
 		this.config = new Config( config, this.constructor.defaultConfig );
-
 		this.config.define( 'plugins', availablePlugins );
+		this.config.define( this._context._getConfigForEditor() );
```
Reinmar: `_getEditorConfig()`?

src/pendingactions.js
```diff
@@ -52,7 +52,7 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  *
  * @extends module:core/plugin~Plugin
```
Reinmar:
```suggestion
 * @extends module:core/contextplugin~ContextPlugin
```

Reinmar: Not sure why the first space is shown by GitHub's diff as removed because it's there in the suggestion ;|

src/contextplugin.js
Reinmar: I also thought that it'd be good to document rules that a context plugin must adhere to. I remember @pjasiun wrote down a couple of them. It's a good place to cover that.

src/plugincollection.js
```diff
@@ -23,37 +23,56 @@ export default class PluginCollection {
...
 		this._availablePlugins = new Map();
 
+		for ( const PluginConstructor of availablePlugins ) {
+			if ( PluginConstructor.pluginName ) {
+				this._availablePlugins.set( PluginConstructor.pluginName, PluginConstructor );
+			}
+		}
+
 		/**
+		 * Map of external plugins which can be retrieved by their constructors or instance.
```
Reinmar: What about retrieving by name?

src/plugincollection.js
Reinmar: Or maybe "contextPlugins" to be more precise? May be easier to grasp.

Reinmar (commented):
I've read through the code and left some comments.

However, the changes are really big and I can't say that I review their consistency. I don't remember many parts of ckeditor5-core already and some changes were made there that I didn't see yet, so it'd be seriously time-consuming. So, I'll schedule a review of this for myself. I guess I'll do that together with polishing the docs. Hopefully, I won't find some major issues.

As for closing this PR – I leave it up to you which improvements proposed by me you want to make now. You can leave the rest to me. What could help, though, would be some summary, what you leave for followups.

Reinmar:
Ticket for followups: https://github.com/ckeditor/ckeditor5/issues/6008

src/plugincollection.js
oskarwrobel: It's an internal map. All loaded context plugins are available under their names.

src/contextplugin.js
oskarwrobel: We decided to keep it where it is.

oskarwrobel:
Ok, I think I've fixed everything 😅

PR #205, opened 2019-12-16
Fix: Made it possible to pass the data processor to be used in one of…

fredck:
… the test utils.

Internal: Made it possible to pass the data processor the one of the test utils. Part or ckeditor/ckeditor5#5977.

### DIFF
--- a/tests/_utils/classictesteditor.js
+++ b/tests/_utils/classictesteditor.js
@@ -33,8 +33,8 @@ export default class ClassicTestEditor extends Editor {
 			this.sourceElement = sourceElementOrData;
 		}
 
-		// Use the HTML data processor in this editor.
-		this.data.processor = new HtmlDataProcessor();
+		// This test util allows for passing the desired data processor through configuration.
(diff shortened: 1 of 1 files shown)


Reinmar:
This adds a config option which will not be documented if it stays like that. There's [https://github.com/ckeditor/ckeditor5/issues/2848](https://github.com/ckeditor/ckeditor5/issues/2848) and it'll soon be 4 years old, but the truth is that we didn't get any major feedback about this. The current method of enabling a DP in a plugin is apparently good enough.

So, here I'd also propose to write a simple function plugin which sets editor.data.processor to the desired value. This way we won't need to introduce a new config option, which, if we'd be doing, must be better thought and documented.

fredck:
> This adds a config option which will not be documented if it stays like that.

Is it a big deal if we're talking about doing so just for a test util, like proposed? (that's really a question :))

fredck:
> Is it a big deal if we're talking about doing so just for a test util, like proposed?

Actually, never mind. I totally agree with you. There are better ways to make this right.

PR #206, opened 2019-12-19
Introduced a factory for classes implementing `Editor`.

### LINKED ISSUE #6015: Core: provide a `Editor` creation factory
Before proposing anything, let me illustrate the two cases I've faced recently, which motivated this issue.

### Case 1

While working on [https://github.com/ckeditor/ckeditor5-markdown-gfm/pull/30](https://github.com/ckeditor/ckeditor5-markdown-gfm/pull/30), I've noticed that, although the plugin is setting `editor.data.processor` to the GFM data processor, the editor classes (classic, inline, etc.) are in any case pointlessly (at that point) setting that same property to the HTML data processor (which later on is overwritten by the plugin).

I figured that it would be safe to set the data processor only after initing all plugins, if no plugin set it. Then, I've also figured that this would involve touching all four official editor classes, basically deleting and pasting the same lines, in the same places. It would not fix other editors out there though.

### Case 2

While coding a custom editor class, it is depressing the amount of copy/paste that has do be done. One does it without even thinking why, copying the boilerplate from the [relative documentation guide](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/ui/custom-editor-creator.html).

All that sounded overcomplicated and pointless. I see this as a weak part of the API.

### Proposal

The creation of a factory that will remove most of the boilerplate from the editor classes and centralize it in a single place.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

fredck:

Feature: Introduced a factory for classes implementing `Editor`. Closes ckeditor/ckeditor5#6015.

---

### Additional information

This PR is paired with 4 other PRs on each of the official editor implementations.

### DIFF
--- a/src/editor/editorfactory.js
+++ b/src/editor/editorfactory.js
@@ -1,0 +1,94 @@
+/**
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import HtmlDataProcessor from '@ckeditor/ckeditor5-engine/src/dataprocessor/htmldataprocessor';
+import getDataFromElement from '@ckeditor/ckeditor5-utils/src/dom/getdatafromelement';
+import attachToForm from './utils/attachtoform';
+import secureSourceElement from './utils/securesourceelement';
+import setDataInElement from '@ckeditor/ckeditor5-utils/src/dom/setdatainelement';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import { isElement } from 'lodash-es';
+
+export default class EditorFactory {
+	create( EditorClass, sourceElementOrData = '', config = {} ) {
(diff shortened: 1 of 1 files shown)


fredck:
Proposing a full implementation for the factory:

*   Includes a factory for `create()`.
*   Includes a factory for `destroy()`.
*   Makes it possible to customize every step of the creation process. [An example can be seen in the connected PR for the ClassicEditor](https://github.com/ckeditor/ckeditor5-editor-classic/pull/101).
*   Reduced the code of all four editor implementations to their very minimum, holding now just code specific for each implementation (no code duplication... kind a... we could make that textarea check a util).

Before moving forward with more tests and documentation, **I'm asking here for feedback. Thanks!**

fredck:
Force-pushed a rebase over v17.0.0.

fredck:
Force-pushed a rebase over v18.0.0.

fredck:
The PR has been moved to ckeditor/ckeditor5#7309, following the CKEditor 5 move to monorepo.

For those interested, be sure to add :+1: and further feedback there.

PR #207, opened 2020-01-03
Fixed tests leaking editor instance / DOM element

mlewand:

Tests: Fixed tests leaking editor instances / DOM elements. See ckeditor/ckeditor5#6002.

---

### Additional information

This PR also adds a set of test utility cleanup methods to be reused by other tests.

For a complete list of PRs see https://github.com/ckeditor/ckeditor5/issues/6002#issuecomment-567834563.

### DIFF
--- a/tests/_utils/cleanup.js
+++ b/tests/_utils/cleanup.js
@@ -17,21 +17,3 @@ export function removeEditorBodyOrphans() {
 		bodyOrphan.remove();
 	}
 }
-
-/**
- * Searches for orphaned editors based on DOM.
- *
- * This is useful if in your tests you have no access to editor, instance because editor
- * creation method doesn't complete in a graceful manner.
- */
-export function destroyEditorOrphans() {
-	const promises = [];
-
-	for ( const editableOrphan of document.querySelectorAll( '.ck-editor__editable' ) ) {
-		if ( editableOrphan.ckeditorInstance ) {
-			promises.push( editableOrphan.ckeditorInstance.destroy() );
-		}
-	}
-
-	return Promise.all( promises );
-}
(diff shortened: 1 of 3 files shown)


tests/_utils/cleanup.js
```diff
@@ -0,0 +1,37 @@
...
+/* global document */
+
+/**
+ * Removes all the `.ck-body` elements available in the DOM.
+ *
+ * It is commonly used to cleanup after editors that test editor crashes.
+ *
+ * See https://github.com/ckeditor/ckeditor5/issues/6018 for more details.
+ */
+export function removeEditorBodyOrphans() {
```
oleq, line 15: I'm not sure we need these `removeEditorBodyOrphans()` and `destroyEditorOrphans()` at all. The first is called once in ckeditor5-core and the latter is called twice in ckeditor5-watchdog. Since they are not required anywhere else, they should be local to the tests that use them.

oleq (commented):
I expressed my only concern here https://github.com/ckeditor/ckeditor5-core/pull/207/files#r363643104.

tests/_utils/cleanup.js
mlewand: You're right about `destroyEditorOrphans`, however `removeEditorBodyOrphans` method is also reused in `editor-*` packages. I believe that having `destroyEditorOrphans` (in `utils/cleanup`) give it some form of discoverability, inlining it in usage place it will get duplicated if someone needs something like that.

@oleq WDYT in this case?

oleq: Alright, that makes sense. I didn't notice `removeEditorBodyOrphans` in `ckeditor5-editor-` because my branches were outdated. Let's ditch `destroyEditorOrphans` only then. 

Let's also have a test for `removeEditorBodyOrphans` maybe?

mlewand: Function `destroyEditorOrphans` was inlined in the watchdog package, while test for `removeEditorBodyOrphans` was added in 4e3c55e.

PR #215, opened 2020-02-13
Introduced copying builtin plugins to get rid of shared references between editors [watchdog]

ma2ciek:

Internal: Introduced copying builtin plugins to get rid of shared references between editors [watchdog]. Part of ckeditor/ckeditor5#6219.

---

### Additional information


Reinmar (changes requested):
Missing tests. Where do we check that the thing is copied?

ma2ciek:
Fixed.

PR #217, opened 2020-02-13
Wait for editor destroy in tests.

### LINKED ISSUE #6139: Some ckeditor5-core tests do not cleanup properly
## 📝 Provide detailed reproduction steps (if any)

Some tests do not wait for `editor.destroy()` promise to resolve:

- https://github.com/ckeditor/ckeditor5-core/blob/8a88cb52c122fd2952f88f8773a03808f32d6037/tests/editor/editor.js#L353-L366
- https://github.com/ckeditor/ckeditor5-core/blob/8a88cb52c122fd2952f88f8773a03808f32d6037/tests/_utils-tests/articlepluginset.js#L43-L45
- and some more

The problem is that if you add asynchronous `afterEach()` some times it never resolves like tests in `DataApiMixin`. It resolves because the `ready` event is not fired (dunno yet why) and the `editor.destroy()` promise does not finish.

I'm worried that this indicates some problems in the API of we're not aware of.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:

Tests: Wait for editor destroy in tests. Closes ckeditor/ckeditor5#6139.

---

### Additional information


mlewand (changes requested):
Overall the change doesn't look bad - but unfortunately, the proposed change causes some watchdog tests to fail:

```
yarn run test -f watchdog/utils/areconnectedthroughproperties.js
```

```
    integration tests
      √ should return false for two different editors
      × should return false for two different editors sharing builtin plugins
      √ should return false for two different editors inheriting default configuration
      × should return false for two different editors sharing builtin plugins

Finished in 0.109 secs / 0.04 secs @ 12:23:33 GMT+0100 (Central European Standard Time)

SUMMARY:
√ 22 tests completed
× 2 tests failed

FAILED TESTS:
  areConnectedThroughProperties()
    integration tests
      × should return false for two different editors sharing builtin plugins
        Chrome 80.0.3987 (Windows 10.0.0)
      AssertionError: expected true to be false

      + expected - actual

      -true
      +false

    at Context.eval (webpack:///./packages/ckeditor5-watchdog/tests/utils/areconnectedthroughproperties.js?:276:128)


      × should return false for two different editors sharing builtin plugins
        Chrome 80.0.3987 (Windows 10.0.0)
      AssertionError: expected true to be false

      + expected - actual

      -true
      +false

    at Context.eval (webpack:///./packages/ckeditor5-watchdog/tests/utils/areconnectedthroughproperties.js?:300:128)
```

jodator:
@mlewand could you double-check this. I don't know why a change in ckeditor5-core tests would cause a problem in ckeditor5-watchdog tests.

Especially if you run ckeditor5 test in isolation and I didn't touch `src/` in this PR.

tests/editor/utils/dataapimixin.js
```diff
@@ -22,10 +22,11 @@ describe( 'DataApiMixin', () => {
 		editor.model.document.createRoot( '$root', 'main' );
 		editor.model.document.createRoot( '$root', 'secondRoot' );
 		editor.model.schema.extend( '$text', { allowIn: '$root' } );
+		editor.fire( 'ready' );
 	} );
 
-	afterEach( () => {
-		editor.destroy();
+	afterEach( async () => {
```
mlewand, line 28: Nit: I see you really love `async` / `await` structure :smiley: Since `editor.destroy()` returns a promise, a shorter way would be to simply `afterEach( () => editor.destroy() );` but it's just a matter of preference 😉

mlewand (approved):
Ok, I didn't sink into this before. Checking again, the reason for the problem was that it required merging latest `master` branch.

LGTM, merged latest `master` and added issue reference for both `editor.fire( 'ready' );` calls.

PR #219, opened 2020-02-24
Added tests/_utils to files entry of package.json

### LINKED ISSUE #6313: Core: publish test utils
The `ckeditor5-core` repo contains many test utils inside the `tests/_utils` directory which are used by many other CKEditor 5 packages in their tests.

The fact that the packages using those utils are working is just an accident, because CKEditor 5 is developed in a monorepo. But, in the moment you try to clone independently any of the CKEditor 5 repos that use the utils, it'll not be possible to run tests anymore.

Additionally, the utils are not exposed for the benefit of any third-party application.

The reason for this is that the [utils are not being included](https://github.com/ckeditor/ckeditor5-core/blob/af60015355280745128beb10f0115b3ca00cd2a1/package.json#L62-L66) when publishing the `ckeditor5-core` package.

## Provide a description of the task

Include `tests/_utils` in the `files` entry of `package.json`.

fredck:

Other: The test/\_utils are not exposed to applications depending on this package. Closes ckeditor/ckeditor5#6313.

---

### Additional information

I wasn't able to test it because it needs a real publishing for that, but it seems to be right.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   "files": [
     "lang",
     "src",
-    "theme"
+    "theme",
+    "tests/_utils"
   ],
   "scripts": {
     "lint": "eslint --quiet '**/*.js'",


pomek:
> I wasn't able to test it because it needs a real publishing for that, but it seems to be right.

Did you try to call `npm pack` in the package directory? It creates an archive that contains files that would be published.

fredck:
> Did you try to call `npm pack` in the package directory? It creates an archive that contains files that would be published.

Thanks for the info. I did this just now and yes, it works like a charm :+1:

Reinmar:
I have a big issue with this :( So far we only published `src/`. We even have `src/dev-utils` in the engine for similar things. Perhaps we should have test utils exposed in the same way if certain utils are useful.

fredck:
Just want to point out that those utils are used extensively in many other CKEditor 5 packages which, if cloned outside the monorepo  repo, get broken.`ckeditor5`

fredck:
Force-pushed a rebase over v18.0.0.

fredck:
The PR has been moved to ckeditor/ckeditor5#7311, following the CKEditor 5 move to monorepo.

For those interested, be sure to add :+1: and further feedback there.

